PR #68, opened 2018-06-02
Twitter Athena example

lindydonna:


README.md
```diff
@@ -80,3 +80,8 @@ Guestbook](https://kubernetes.io/docs/tutorials/stateless-application/guestbook/
 #### [AWS Voting App with Containers](cloud-ts-voting-app/)
 
 A simple voting app that uses Redis for a data store and a Python Flask app for the frontend, demonstrating the high-level framework `@pulumi/cloud`.
+
+#### [AWS Athena Twitter Analyzer](cloud-js-twitter-athena/)
```
joeduffy: Should we change this to be a table, more like https://github.com/serverless/examples?

lindydonna: That would be a great idea. I filed https://github.com/pulumi/examples/issues/76.

Is this example good to merge as-is?

PR #74, opened 2018-06-07
Add api example using @pulumi/aws-serverless

ellismg:
This applications behaves the same as our `cloud-aws-httpendpoint`
example, but uses the newer @pulumi/aws-serverless framework.

### DIFF
--- a/aws-serverless-js-httpendpoint/index.js
+++ b/aws-serverless-js-httpendpoint/index.js
@@ -1,0 +1,62 @@
+const pulumi = require("@pulumi/pulumi")
+const aws = require("@pulumi/aws")
+const serverless = require("@pulumi/aws-serverless");
+
+// A table to hold the counts of each route that is hit. We use the route name
+// as the key for the the table.  The read and write capacity are intentionally
+// small for this toy example.  The AWS Free Teir provides 25 units of read and
+// write capacity across all DynamoDB Tables in your account.
+const table = new aws.dynamodb.Table("counterTable", {
+    attributes: [{
+        name: "route",
+        type: "S",
+    }],
+    hashKey: "route",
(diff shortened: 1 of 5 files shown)


aws-serverless-js-httpendpoint/index.js
```diff
@@ -0,0 +1,62 @@
...
+        Key: { route }
+    }).promise()).Item;
+
+    let count = (value && value.count) || 0;
+
+    // Increment the count and write it back to DynomoDB.
+    await (dynomo.put({
+        TableName: tableName,
+        Item: { route, count: ++count }
+    })).promise();
```
lukehoban: Since we are using Dynamo directly, we can simplify and improve the above by using an atomic increment instead of get/set.  See: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html#WorkingWithItems.AtomicCounters

aws-serverless-js-httpendpoint/index.js
```diff
@@ -0,0 +1,62 @@
...
+        name: "route",
+        type: "S",
+    }],
+    hashKey: "route",
+    readCapacity: 1,
+    writeCapacity: 1,
+});
+
+// The handler for GET /{route+}.
+const getHandler = async (event) => {
```
lukehoban: May as well make this `async function getHandler(event) { ... }`.

lukehoban: In our other JS examples, we recently decided to avoid `async/await` and use promises directly instead, so that examples will work on earlier Node versions.  That might be a little more painful here, but probably worth being consistent.

lindydonna:
@ellismg Is this PR still in progress?

PR #81, opened 2018-06-15
Add Kubernetes TypeScript Sock Shop example

hausdorff:


kubernetes-ts-sock-shop/Pulumi.yaml
```diff
@@ -0,0 +1,4 @@
+name: sock-shop
+runtime: nodejs
+description: The standard Weave Works Sock Shop demo based on https://github.com/microservices-demo/microservices-demo
```
ericrudder: Weaveworks

PR #88, opened 2018-06-18
Add azure-ts-appservice example

mikhailshilkov:


azure-ts-appservice/index.ts
```diff
@@ -0,0 +1,104 @@
...
+});
+
+const storageContainer = new azure.storage.Container(`${name}-c`, {
+    resourceGroupName: resourceGroup.name,
+    storageAccountName: storageAccount.name,
+    containerAccessType: "private",
+});
+
+const map: pulumi.asset.AssetMap = {};
+map["index.html"] = new pulumi.asset.StringAsset("<html><body><h1>Greetings from Azure App Service!</h1></body></html>");
```
lukehoban: Was there any reason this could not be written as:

```typescript
const map: pulumi.asset.AssetMap = {
    "index.html": new pulumi.asset.StringAsset("<html><body><h1>Greetings from Azure App Service!</h1></body></html>"),
};
```

azure-ts-appservice/index.ts
```diff
@@ -0,0 +1,104 @@
...
+        "WEBSITE_RUN_FROM_ZIP": codeBlobUrl,
+        "ApplicationInsights:InstrumentationKey": appInsights.instrumentationKey,
+        "APPINSIGHTS_INSTRUMENTATIONKEY": appInsights.instrumentationKey
+    },
+
+    connectionStrings: [{
+        name: "db",
+        value: sqlServer.name.apply(
+            server => database.name.apply(
+                db => `Server=tcp:${server}.database.windows.net;initial catalog=${db};user ID=${username};password=${pwd};Min Pool Size=0;Max Pool Size=30;Persist Security Info=true;`)),
```
lukehoban: There is a `pulumi.all` function which can be useful in cases like this:

```typescript
pulumi.all([sqlServer, database]).apply(([server, db]) => 
    `Server=tcp:${server}.database.windows.net;initial catalog=${db};user ID=${username};password=${pwd};Min Pool Size=0;Max Pool Size=30;Persist Security Info=true;`
);
```

azure-ts-appservice/index.ts
```diff
@@ -0,0 +1,104 @@
...
+        type: "SQLAzure"
+    }]    
+});
+
+exports.endpoint = app.defaultSiteHostname.apply(n => `https://${n}`);
```
lukehoban: Nit: Trailing newline.

azure-ts-appservice/sas.ts
```diff
@@ -0,0 +1,37 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as azure from "@pulumi/azure";
+
+import * as azurestorage from "azure-storage";
+
+// Given an Azure blob, create a SAS URL that can read it.
+export function signedBlobReadUrl(
```
lukehoban: We'll likely want to move this code out into `@pulumi/azure` itself as a helper since it appears to be needed broadly to support App Service and Functions.  I've opened https://github.com/pulumi/pulumi-azure/issues/70 to track this, which we can tackle separately.

azure-ts-appservice/README.md
```diff
@@ -0,0 +1,62 @@
...
+1.  Build the Typescript project:
+
+    ```
+    $ npm run build
+    ```
+
+1. Define SQL Server password (make it complex enough to satisfy Azure policy):
+
+    ```
+    pulumi config set azpulumi:sqlPassword <value>
```
lukehoban: If you do take the suggestion below to replace `azpulumi` with `pulumi.getStack()`, then this can become just `pulumi config set sqlPassword <value>`.

azure-ts-appservice/index.ts
```diff
@@ -0,0 +1,104 @@
...
+map["index.html"] = new pulumi.asset.StringAsset("<html><body><h1>Greetings from Azure App Service!</h1></body></html>");
+
+const blob = new azure.storage.ZipBlob(`${name}-b`, {
+    resourceGroupName: resourceGroup.name,
+    storageAccountName: storageAccount.name,
+    storageContainerName: storageContainer.name,
+    type: "block",
+
+    content: new pulumi.asset.AssetArchive(map)
+             // to pack a folder use new pulumi.asset.FileArchive("folder")
```
lukehoban: Is it worth doing the folder mode by default here?  I assume that will be more "normal".

azure-ts-appservice/index.ts
mikhailshilkov: I copy-pasted it from `azure-ts-function` sample, I'll change

lukehoban: Got it :smile:.  We'll update the code in the other example as well.

azure-ts-appservice/index.ts
mikhailshilkov: In my real example I had an ASP.NET Core app there, but wasn't sure just making a folder with `index.html` would make sense

azure-ts-appservice/README.md
mikhailshilkov: Hmm... `getStack()` returns stack name e.g. `mystack` while default config is saved under app name e.g. `azure-appservice`

azure-ts-appservice/README.md
```diff
@@ -34,7 +34,7 @@ with App Service.
 1. Define SQL Server password (make it complex enough to satisfy Azure policy):
 
     ```
-    pulumi config set azpulumi:sqlPassword <value>
+    pulumi config set <stack>:sqlPassword <value>
```
lukehoban: You can leave the `<stack>:` off, and Pulumi will default to a config variable for the current stack - `pulumi config set sqlPassword <value>`

azure-ts-appservice/index.ts
lukehoban: Yeah - I think a folder with just an `index.html` makes sense - and is closer to what most folks (including yourself!) would want to do here.

This pattern of building up the archive in memory appears to be borrowed from the Functions example, but was only used there because that code actually is dynamically constructing the contents to include in the upload as part of the program.  In the case of App Service here, it would be very rare I expect to need/want to do that.  This will also remove some complexity here.

azure-ts-appservice/README.md
mikhailshilkov: As I said, it seems that default is project name not stack name. I got error at least.

lukehoban:
@mikhailshilkov Awesome!  And thanks for the PR.  I've gone ahead and merged this in.

PR #91, opened 2018-06-24
Add azure-python-webserver example.

scottmckenzie:


azure-python-webserver/__main__.py
```diff
@@ -0,0 +1,92 @@
...
+        'sku': '16.04-LTS',
+        'version': 'latest',
+    },
+    boot_diagnostics={
+        'enabled': True,
+        'storage_uri': sa.primary_blob_endpoint,
+    }
+)
+
+# Note - due to a bug in the terraform-provider-azurerm, the public IP address is not yet populated corerctly.
```
swgillespie: Do you have any details on this bug? Would be awesome if we could fix this.

azure-python-webserver/README.md
```diff
@@ -0,0 +1,56 @@
...
+    ```
+
+1.  Check the IP address:
+
+    ```
+    $ pulumi stack output private_ip
+    10.0.2.4
+    ```
+
+*TODO*: Expose the Public IP address as well so that the VM can be SSH'd into or CURL'd directly.
```
swgillespie: I'm assuming this is because of the `terraform-provider-azuream` bug you mentioned?

azure-python-webserver/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: azure-python-webserver
+description:
```
swgillespie: Can you add a short description here?

azure-python-webserver/__main__.py
lukehoban: I believe this text was copied from the JS version of this example.  It is ultimately due to the issue outlined in https://github.com/terraform-providers/terraform-provider-azurerm/issues/764#issuecomment-365019882.

Since this won't work currently - I'd suggest we just leave this line off.  It's possible to use `pulumi refresh` or some other tricks to work around this, but probably not necessary to focus on for this simple example.

lukehoban: I just submitted https://github.com/pulumi/examples/pull/93 with a fix for this in JavaScript.  Should be able to do similar in Python, but likely even simpler due to Python's synchronous model for resource creation.

stack72:
I am going to close this out - it's been almost a year since this was opened as a PR and feedback was given. We currently have https://github.com/pulumi/examples/tree/master/azure-py-webserver - if you feel this needs further work, please do comment below

I want to thank you for your contribution! 

Paul

PR #104, opened 2018-07-21
Updates to Jenkins

ejsmith:
Use name for PVC name to avoid conflicts.
Use v1 Deployment.
Allow storage class to be configured.
Use name from args so it doesn't need to be specified twice.


joeduffy:
Nice! @swgillespie can you take a deeper look?

PR #106, opened 2018-07-28
Add an AWS EC2 Ruby on Rails example

joeduffy:
This ports the official AWS CloudFormation Application Framework
template for a Ruby on Rails server that uses EC2 and a MySQL local
database. I've implemented some experimental support for
AWS::CloudFormation::Init-like config-sets, that install files and
packages, and run scripts and services, on the VM during startup.


aws-ts-ruby-on-rails/ami.ts
```diff
@@ -0,0 +1,95 @@
+import * as aws from "@pulumi/aws";
+
+const instanceTypeToArch: {[instanceType: string]: string} = {
+    "t1.micro"    : "PV64" ,
```
lukehoban: FWIW- this list is missing about half the EC2 instance types.

aws-ts-ruby-on-rails/init.ts
```diff
@@ -0,0 +1,112 @@
...
+type InitServices = {[service: string]: InitService};
+
+interface InitService {
+    enabled?: boolean;
+    ensureRunning?: boolean;
+}
+
+// createUserData produces a single user data script out of the given init structure.
+export async function createUserData(configs: string[], init: Init): Promise<string> {
+    let script = `#!/bin/bash -xe\n\n`;
```
lukehoban: This appears to be taking a cloud-init like schematic as input and lowering it to bash. You should be able to just generate cloud-init format directly and pass that as userData.

aws-ts-ruby-on-rails/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "scripts": {
+        "build": "tsc"
+    },
+    "devDependencies": {
+        "typescript": "^2.7.2"
+    },
+    "dependencies": {
+        "@pulumi/aws": "^0.14.0",
+        "@pulumi/pulumi": "^0.14.0",
+        "@types/mustache": "^0.8.31",
```
lukehoban: Technically @types can be devDependencies.

aws-ts-ruby-on-rails/ami.ts
joeduffy: Eek! I took this straight from the most recent Application Template in the AWS Console!

I'd happily replace this with a call to some standard implementation of this. Should we just create a simple unofficial package for this? Or, do we prefer using the dynamic querying variant that I blogged about recently.

I was trying to keep this as faithful a port as possible, and am not keen to get into the business of manually maintaining each and every list here, but I agree it's a shame if people copy this and it doesn't work for all possible regions.

aws-ts-ruby-on-rails/init.ts
joeduffy: I spent a little time looking into this. Unless I'm missing something, cloud-init isn't an identical 1-for-1 match. It does support all the same capabilities, however (minus the package management commands, again unless I'm missing some extensibility point (which is likely)).

My goal was to make it easy to convert existing CloudFormation. For that reason, I wonder if it makes sense to support the types that I have in here, and simply lower to cloud-init. There are several reasons why doing cloud-init directly actually kind of stinks: the MIME multi-part encoding required for multi-file configs, and the odd merging behavior it has for duplicate dictionaries (this example is a good one where you want to have multiple conceptual steps).

I agree, lowering to Bash is as gross as it gets, so cloud-init seems better.

aws-ts-ruby-on-rails/package.json
joeduffy: Good catch, will fix.

PR #114, opened 2018-08-14
Add an AKS/Helm example

joeduffy:
This example deploys a new Azure Kubernetes Service (AKS) cluster and
then also deploys an Apache Helm Chart into that cluster. It
automatically provisions an Active Directory service principal to use
for the cluster's auth.

### DIFF
--- a/azure-ts-aks-helm/index.ts
+++ b/azure-ts-aks-helm/index.ts
@@ -1,0 +1,20 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as azure from "@pulumi/azure";
+import * as helm from "@pulumi/kubernetes/helm";
+import * as k8s from "@pulumi/kubernetes";
+import * as pulumi from "@pulumi/pulumi";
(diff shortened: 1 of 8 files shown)


joeduffy:
@hausdorff @pgavlin Can one of you take a look at this?

azure-ts-aks-helm/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const apache = new helm.v2.Chart("apache", {
+    repo: "bitnami",
+    chart: "apache",
+    version: "1.0.0",
+}, { providers: { kubernetes: k8sProvider } });
+
+export let cluster = k8sCluster.name;
+export let kubeConfig = k8sCluster.kubeConfigRaw;
+export let serviceIP =
```
pgavlin: Does this actually work?

joeduffy: It will once @hausdorff fixes https://github.com/pulumi/pulumi-kubernetes/issues/137 today.

PR #127, opened 2018-08-31
Simplify the Kubernetes Guestbook Example

joeduffy:
This change makes a variant of the Kubernetes Guestbook Example
that shows off using classes to reduce boilerplate. It is slightly
opinionated, so may not be perfect for all use cases, but can at
least be shown to stimulate thought about what's possible.

As part of this, I've reverted the existing Guestbook Example to
be as straight a port as possible, without any abstraction. These
can now be easily compared as a before/after pair.

Of course, until we figure out where we're landing on KubeJS, etc.,
we won't know the shapes here, but again it will help tell the
story of why Pulumi is unique and advantageous in the meantime.


joeduffy:
@hausdorff I'm going to merge, would love your feedback if you have the time!

PR #129, opened 2018-09-08
Add a Kubernetes stateless app deployment example

joeduffy:
This is a straight port of
https://kubernetes.io/docs/tasks/run-application/run-stateless-application-deployment/.
We will be turning this into a tutorial on the pulumi.io website.

### DIFF
--- a/kubernetes-ts-nginx/Pulumi.yaml
+++ b/kubernetes-ts-nginx/Pulumi.yaml
@@ -1,0 +1,3 @@
+name: k8s-nginx
+runtime: nodejs
+description: Example of a Kubernetes Stateless Application Deployment, using Nginx.
--- a/kubernetes-ts-nginx/README.md
+++ b/kubernetes-ts-nginx/README.md
@@ -1,0 +1,70 @@
+# Run a Stateless Application Using a Deployment
+
+A version of the [Kubernetes Stateless Application Deployment](
+https://kubernetes.io/docs/tasks/run-application/run-stateless-application-deployment/) example that uses Pulumi.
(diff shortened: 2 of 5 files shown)


octocruise:
i wonder if there's an opportunity for some activity with the Docker/Kube playground ... it eliminates a lot of the pre-work to get a cluster going.

have you looked at this stuff at all?  (they are linked in the orig "inspiration" page)

in any case, even if you hate the idea of working with Docker direcly (too hard to partnew, or we're not "pur enough" etc), we might find other inspiration from the docker "classroom" stuff .... while the docker sites (classroom, training) are ugly, there are some ideas wanting to bust out.


  1.  we can leverage the idea of the "classroom" to optimize the landing pages for training.  i know you've played with this a lot, esp. in our flow, however, i think there might be some more ideas buried in here.  (again, the execution is ugly, however, the ideas are there.)
  2.  they tend to organize into "basics, advanced, production" ... the terms change, etc, however, we can possibly adopt the third category and see from metrics who goes deepest (maybe tracking a log or debug, etc.) ... i think at a minimum, we should be deliberate about where we want the "working with teams" (new, unwritten stuff) to go.
  3.  they use the right hand gutter to "advertise" other offerings.  we could use this as a feed for "where we will be" (events, etc -- hit us up at Kubecon) and possibly offer training ... we can also "beg" again for email/twitter/etc
  4.  they mix in videos in the flow; we can add asciinema or vids in a few places too.  i'm still amazed that devs really want this rather than "reading" however, the numbers sadly continue to show this ... we could add an explicit goal of "one new animation per milestone" or something like that.

i know you think a lot about 1st impression which is why i mention some of this stuff.


________________________________

PR #137, opened 2018-09-18
Small fixes to several Kubernetes examples

hausdorff:


azure-ts-aks-helm/index.ts
```diff
@@ -4,14 +4,18 @@ import * as helm from "@pulumi/kubernetes/helm";
 import * as k8s from "@pulumi/kubernetes";
 import { k8sCluster, k8sProvider } from "./cluster";
 
-const apache = new helm.v2.Chart("apache", {
-    repo: "bitnami",
-    chart: "apache",
-    version: "1.0.0",
-}, { providers: { kubernetes: k8sProvider } });
+const apache = new helm.v2.Chart(
```
lukehoban: Any particular reason for this layout?  We’re reasonably consistent on formatting for this style of thing and typically use the original format.

hausdorff: I switch to using the TypeScript formatter, and that's how it spit it out.

hausdorff: I'm going to interpret the LGTM as non-blocking comments. My recommendation is that we all transition to a TypeScript formatter and just use that.

PR #140, opened 2018-09-20
Re-add test for aws-py-webserver

lukehoban:


mikhailshilkov:
@lukehoban Do you remember what was the status of this PR? Is it worth reviving it?

lukehoban:
Let’s close for now. We’ll want to come back and re-add tests for many of the examples - but this PR appears to be stale.

PR #141, opened 2018-09-20
Example test added for azure-ts-functions.

daveremy:


misc/test/Gopkg.lock
```diff
@@ -3,26 +3,33 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:6978a38432a017763a148afbc7ce6491734b54292af7d3e969d84d2e9dd242e2"
```
lukehoban: It looks like you are using a more recent `dep` than what we are using currently across Pulumi projects.  See https://github.com/pulumi/scripts/blob/master/ci/install-common-toolchain.sh for versions.  (I may have misled you in person yesterday pointing at https://github.com/pulumi/pulumi#development, which may pick up a newer version than we currently support).

misc/test/examples_test.go
```diff
@@ -67,6 +67,32 @@ func TestExamples(t *testing.T) {
 				expectHelloWorld(t, stack.Outputs["publicIP"])
 			},
 		}),
+		base.With(integration.ProgramTestOptions{
+			Dir:       path.Join(cwd, "..", "..", "azure-ts-functions"),
```
lukehoban: I already submitted test coverage for this one yesterday in https://github.com/pulumi/examples/pull/139 based on the work we did in the AM.

misc/test/examples_test.go
```diff
@@ -67,6 +67,32 @@ func TestExamples(t *testing.T) {
...
+					if !assert.NoError(t, err) {
+						return false
+					}
+					defer resp.Body.Close()
+					body, err := ioutil.ReadAll(resp.Body)
+					if !assert.NoError(t, err) {
+						return false
+					}
+					return assert.Contains(t, string(body), "Greetings from Azure Functions!")
+				}(t, stack.Outputs["endpoint"])
```
lukehoban: Along the lines of #139, let's try to share more of the HTTP request functionality across tests and generalize that as needed as we expand test coverage.

daveremy:
Thanks Luke, your comments make sense.  I missed that you had merged, that refactor on the httprequest looks great.  I'll close this.

PR #145, opened 2018-09-30
azure-ts-aks-mean example test

daveremy:
To make it a bit better to run a test directly defaulted the two required environment variables (AWS_REGION, ARM_ENVIRONMENT).

Implement the azure-ts-aks-mean test.  Had to reduce the size of the DNS
Prefix because the derived name was larger than 45 chars.

### DIFF
--- a/azure-ts-aks-mean/README.md
+++ b/azure-ts-aks-mean/README.md
@@ -51,7 +51,7 @@ npm install
 
     > **Note**: Due to an issue in the Azure Terraform Provider (https://github.com/terraform-providers/terraform-provider-azurerm/issues/1635) the
     > creation of an Azure Service Principal, which is needed to create the Kubernetes cluster (see cluster.ts), is delayed and may not 
-    > be available when the cluster is created.  If you get a Service Principal not found error, as a work around, you should be able to run `Pulumi up`
+    > be available when the cluster is created.  If you get a Service Principal not found error, as a work around, you should be able to run `pulumi up`
     > again, at which time the Service Principal should have been created.
 
     ```sh
--- a/azure-ts-aks-mean/README.md
+++ b/azure-ts-aks-mean/README.md
@@ -49,6 +49,11 @@ npm install
 
 1. Perform the deployment:
 
+    > **Note**: Due to an issue in the Azure Terraform Provider (https://github.com/terraform-providers/terraform-provider-azurerm/issues/1635) the
+    > creation of an Azure Service Principal, which is needed to create the Kubernetes cluster (see cluster.ts), is delayed and may not 
+    > be available when the cluster is created.  If you get a Service Principal not found error, as a work around, you should be able to run `Pulumi up`
+    > again, at which time the Service Principal should have been created.
+
     ```sh
     $ pulumi up
     Updating stack 'azure-mean'
--- a/misc/test/examples_test.go
+++ b/misc/test/examples_test.go
@@ -19,11 +19,13 @@ import (
 func TestExamples(t *testing.T) {
 	awsRegion := os.Getenv("AWS_REGION")
 	if awsRegion == "" {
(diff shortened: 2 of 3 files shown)


lukehoban:
Looks like this is hitting https://github.com/terraform-providers/terraform-provider-azurerm/issues/1635.  It may not be possible to work around that in the test here until they underlying bug is fixed.  Users of this example will (in the meantime), likely need to do the update - get this failure - and then try the update again and it will succeed.  Possibly worth noting in the README for this example, and then commenting out this test with a TODO pointing to that issue.

daveremy:
> Looks like this is hitting [terraform-providers/terraform-provider-azurerm#1635](https://github.com/terraform-providers/terraform-provider-azurerm/issues/1635). It may not be possible to work around that in the test here until they underlying bug is fixed. Users of this example will (in the meantime), likely need to do the update - get this failure - and then try the update again and it will succeed. Possibly worth noting in the README for this example, and then commenting out this test with a TODO pointing to that issue.

Ah, I ran into this issue when running the example, but, of course, when I reran it finished successfully.  What about the possibility of changing the example to have the service principle passed in and showing in the readme how to create a service principal using the azure cli?  I assume that would work.  I'm not sure whether/how that would be done with the test runner here though ...

lukehoban:
> I assume that would work. I'm not sure whether/how that would be done with the test runner here though ...

Right, not sure that would help with the test - and likely makes the example net harder to use. Suggest just a comment in readme and TODO pointing to issue in the tests.

daveremy:
One thing that seems strange is that the azure-ts-aks-helm example test uses exactly the same code in cluster.ts where the k8s cluster is created ... yet it seems to finish consistently ... it runs quite a bit longer, perhaps there is more time for the sp to show up.

daveremy:
For some reason CI was failing on this for multiple runs but is now passing.  Nothing in the change should affect the run as far as i can tell.  @lukehoban (or others) you down with introducing this skipped tests concept?  Let me know, I like the idea that skipped tests have some sort of nag, but also ok with simply commenting it out.

lukehoban:
If these tests can't run reliably, we should comment them out with a TODO pointing at terraform-providers/terraform-provider-azurerm#1635.

azure-ts-aks-mean/README.md
```diff
@@ -49,6 +49,11 @@ npm install
 
 1. Perform the deployment:
 
+    > **Note**: Due to an issue in the Azure Terraform Provider (https://github.com/terraform-providers/terraform-provider-azurerm/issues/1635) the
+    > creation of an Azure Service Principal, which is needed to create the Kubernetes cluster (see cluster.ts), is delayed and may not 
+    > be available when the cluster is created.  If you get a Service Principal not found error, as a work around, you should be able to run `Pulumi up`
```
lukehoban: `pulumi up`

misc/test/examples_test.go
```diff
@@ -103,20 +105,52 @@ func TestExamples(t *testing.T) {
...
+				"aws:region": awsRegion,
+			},
+			ExtraRuntimeValidation: func(t *testing.T, stack integration.RuntimeValidationStackInfo) {
+				assertHTTPHelloWorld(t, stack.Outputs["public_dns"])
+			},
+		}),
+	}
+
+	// Print skipped example tests to the test output
+	if len(skippedExamples) > 0 {
```
lukehoban: I don't think we need this - just commented with a TODO is fine.

azure-ts-aks-mean/README.md
daveremy: thx, fixed.

misc/test/examples_test.go
daveremy: got it.  backed it out, moved to comment and TODO

PR #148, opened 2018-10-02
Test for azure-ts-appservice and test fixes

daveremy:


azure-ts-appservice/README.md
```diff
@@ -28,10 +28,9 @@ with App Service.
 1. Define SQL Server password (make it complex enough to satisfy Azure policy):
 
     ```
-    pulumi config set <stack>:sqlPassword <value>
+    pulumi config set --secret <stack>:sqlPassword <value>
```
lukehoban: Note that we should be able to remove the `<stack>` part here as well.

azure-ts-appservice/README.md
```diff
@@ -50,7 +49,11 @@ with App Service.
 
     ```
     $ pulumi stack output endpoint
-    https://azpulumi-as0ef47193.azurewebsites.net
+    https://azpulumi-as0ef47193.azurewebsites.net (you will get a different url)
```
lukehoban: Don't think we have to add this here (we have similar many other places as well).

azure-ts-appservice/README.md
```diff
@@ -50,7 +49,11 @@ with App Service.
...
     $ curl "$(pulumi stack output endpoint)"
-    Greetings from Azure App Service!
+    <html>
+        <body>
+            <h1>Greetings from Azure App Service!</h1>
+        </body>
+    </html>%
```
lukehoban: What's the `%` for?

azure-ts-appservice/README.md
daveremy: it needs to be `pulumi config set --secret azure-appservice:sqlPassword <value>` right?

azure-ts-appservice/README.md
daveremy: cut and paste issue probably

azure-ts-appservice/README.md
lukehoban: You can leave that off and it will default to the project-scoped config.

PR #157, opened 2018-10-20
Add StatefulSet example

hausdorff:


kubernetes-ts-helm-wordpress/README.md
```diff
@@ -46,7 +46,17 @@ $ pulumi stack init
...
+This chart requires you to set two configuration values: `wordpressPassword`, `mariadbPassword`, and
+`mariadbRootPassword`. If these values are not set, Helm will re-generate the passwords each
+upgrade, which will cause the upgrade to fail. (See
+[helm/charts#5167](https://github.com/helm/charts/issues/5167).)
+
+```sh
+p config set isMinikube false # This chart disallows minikube.
+p config set wordpressPassword <value>
+p config set mariadbPassword <value>
+p config set mariadbRootPassword <value>
```
lukehoban: This isn't a standard alias - we should use `pulumi config` here.

kubernetes-ts-stateful/cassandra.ts
```diff
@@ -0,0 +1,113 @@
...
+    transforms?: [(a: inputApi.apps.v1.StatefulSet) => void];
+}
+
+export class Cluster extends k8s.apps.v1.StatefulSet {
+    constructor(name: string, args?: ClusterOpts, opts?: pulumi.CustomResourceOptions) {
+        super(name, base(args), opts);
+    }
+}
+
+function base(opts?: ClusterOpts): inputApi.apps.v1.StatefulSet {
```
lukehoban: Could `base` have a nicer name?

kubernetes-ts-stateful/cassandra.ts
```diff
@@ -0,0 +1,113 @@
...
+    resourceRequest?: pulumi.Input<object>;
+    securityCapabilities?: pulumi.Input<inputApi.core.v1.Capabilities>;
+    heapSize?: pulumi.Input<string>;
+    heapNewSize?: pulumi.Input<string>;
+    volumeClaimRequests?: pulumi.Input<object>;
+
+    transforms?: [(a: inputApi.apps.v1.StatefulSet) => void];
+}
+
+export class Cluster extends k8s.apps.v1.StatefulSet {
```
lukehoban: Perhaps a comment on this?

kubernetes-ts-stateful/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "k8s-nginx",
+  "version": "0.1.0",
+  "devDependencies": {
+    "@types/node": "^9.3.0"
+  },
+  "dependencies": {
+    "@pulumi/kubernetes": "^0.16.0",
+    "@pulumi/pulumi": "^0.15.0"
```
lukehoban: Move this to `0.16.0`, and maybe Kubernetes to `0.17.0`?

ellismg: We probably just want to be using `latest` here, as with our other examples right?  That way we don't have to keep updating these examples every time we bump (minor) package versions?

mikhailshilkov:
@hausdorff Do you remember why this was never finished? Worth adding last bits and merge?

justinvp:
@hausdorff, what's the status of this PR?

hausdorff:
This PR is effectively abandoned. We originally started this when people were asking for `StatefulSet` examples, but I don't think it's particularly well-written, and I don't think it elucidates anything that's particularly mysterious.

If there is demand for it I can fix it up but this is not something that I think is particularly high leverage. LMK what you think.

justinvp:
> This PR is effectively abandoned.

OK, let's close the PR then.

stack72:
Closed as abandoned - thanks for clearing this up @hausdorff

PR #163, opened 2018-10-30
WIP - Add Pulumi button to examples

justinvp:
Before I go crazy updating all the examples, looking for feedback on updating this first example.

You can actually try out the Pulumi button, by [viewing the README.md](url
https://github.com/pulumi/examples/blob/5a6945720233dcb376966e34860170d2d9f54700/aws-js-s3-folder/README.md) and clicking the button.

In this case, I'm updating the instructions holistically to include the button as part of "Getting Started".

Alternatively, to start, I could just go through and add the button to the top of all the READMEs, and then do a more holistic approach of integrating it into the instructions for each example subsequently.


aws-js-s3-folder/README.md
```diff
@@ -3,50 +3,29 @@
...
-    ```
+After creating the project, a command will be provided (e.g. `pulumi new https://github.com/pulumi/examples/aws-js-s3-folder -s <owner>/<stack-name>`) that you can run to get the source code and preview an initial deployment.
 
-1.  Set the AWS region:
+When you run the command, you'll be presented a preview of the AWS resources that will be created.  Once you've reviewed and selected "Yes", the deployment will begin.
 
-    ```
-    $ pulumi config set aws:region us-west-2
-    ```
+## Getting Started (Command Line)
```
justinvp: I'm thinking about cutting this section. On the one hand, it's nice to show the single command that you can run to get started. On the other hand, it just adds more words to the README, making it seemingly more complex, when showing just the button would be sufficient. Thoughts?

justinvp:
Replaced by #167

PR #178, opened 2018-11-13
Add multi-cloud example targeting AWS and GCP

### LINKED ISSUE #177: Add a multi-cloud example
We should add an example showing provisioning across two or more different clouds.

jen20:
This is something that came up in the [community chat][chat].

Fixes #177.

[chat]: https://pulumi-community.slack.com/archives/C84L4E3N1/p1542089227713300


multicloud-ts-buckets/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "aws-typescript",
+    "devDependencies": {
+        "@types/node": "latest"
+    },
+    "dependencies": {
+        "@pulumi/aws": "latest",
+        "@pulumi/gcp": "^0.16.1",
```
lukehoban: Latest?

multicloud-ts-buckets/README.md
```diff
@@ -0,0 +1,67 @@
+# Multi-Cloud in AWS and GCP using TypeScript
```
lukehoban: We recently added a button to each example to automatically “new” it. We should add that here as well (it will likely also require adding some default config in Pulumi.yaml).

multicloud-ts-buckets/package.json
jen20: I’ve changed this to latest.

multicloud-ts-buckets/README.md
jen20: Is it ok to have configuration items without a default value? It doesn’t seem like there is an acceptable default for GCP project.

jen20: Looks like the other GCP examples list project in the template with no default, which is what I have done here too.

PR #180, opened 2018-11-14
Update Python examples to support Pulumi >= 0.16.4

### LINKED ISSUE #179: Update Python examples to work with Python3
Current Python examples will not work with Pulumi >=0.16.4.

lukehoban:
Fixes #179.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 bin/
 node_modules/
 *.pyc
+.Python
+include/
+lib/
 yarn.lock
 package-lock.json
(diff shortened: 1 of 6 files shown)


yapith-suplari:
@lukehoban copied from slack #python channel:
```
Sometimes your Stepfunction instance consists of multiple lambda steps, I'm not sure the apply API is intuitive in building multi-step definition.

In my case, the steps are built programmatically before they are aggregated into a single stepfunction definition. For my understanding, what was wrong with the previous approach with string-based definition?
```

Issue #156, opened 2018-10-20
Create an example showcasing new aws package serverless


joeduffy:
I don't think we hav any examples showcasing the new serverless functionality from the AWS package. We have https://github.com/pulumi/pulumi-aws/blob/master/examples/stream/index.ts in the AWS repo, which would be a fine start, but something a little more complete would be nice.

PR #181, opened 2018-11-14
Add new examples

CyrusNajmabadi:
Fixes https://github.com/pulumi/examples/issues/156


aws-ts-twitter-athena/index.ts
```diff
@@ -0,0 +1,123 @@
+import * as pulumi from "@pulumi/pulumi";
```
lukehoban: Should we just delete `cloud-js-twitter-athena` now?  It was explicitly AWS-specific, and didn't take advantage of anything that is unique to `cloud` - so I see no reason we wouldn't prefer folks to look at this example instead of that one.

misc/test/examples_test.go
```diff
@@ -118,6 +118,20 @@ func TestExamples(t *testing.T) {
...
+				"aws:region":     awsRegion,
+			},
+			ExtraRuntimeValidation: func(t *testing.T, stack integration.RuntimeValidationStackInfo) {
+				maxWait := 10 * time.Minute
+				endpoint := stack.Outputs["endpoint"].(string)
+				assertHTTPResultWithRetry(t, endpoint + "hello", maxWait, func(body string) bool {
+					return assert.Contains(t, body, "route")
+				})
+			},
+		}),
```
lukehoban: :+1: for test coverage!

aws-ts-ruby-on-rails/index.ts
```diff
@@ -15,7 +15,7 @@ const webSg = new aws.ec2.SecurityGroup("webServerSecurityGroup", {
 });
 
 const webServer = new aws.ec2.Instance("webServer", {
-    ami: getLinuxAmi(config.instanceType),
+    ami: getLinuxAmi(<any>config.instanceType),
```
lukehoban: This is a shame for two reasons:
1. It points out that adding to strongly typed enums like this will break components which use them but aren't yet updated.
2. Turns out the implementation of `getLinuxAmi` doesn't even use this parameter - so this really should just be removed.

PR #184, opened 2018-11-20
WIP: Add Istio example

hausdorff:


mikhailshilkov:
@hausdorff What still needs to be done here?

stack72:
Going to close this out - we can revisit this if we need to

SRodi:
@hausdorff Any plans to reopen this and configure it to work with Istio version >1.6?
It would be nice to be able to provision Istio with Pulumi. I am aware this is already possible on GKE with Istio plugin, but it would be useful to have the full Istio installation which can be provisioned also on different Clouds including EKS, AKS etc. 
Anybody has suggestions on other/better ways to automate the provisioning of Istio?

hausdorff:
@SRodi apologies for the delays—I left Pulumi. You should still be able to use Pulumi through the [Helm integration](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/kubernetes/helm/v3/)! Helm is still the official distribution channel of Istio, as far as I know.

cc @lblackstone who I think owns these efforts since I left.

Issue #186, opened 2018-11-28
Update examples to use "latest" instead of "dev"


justinvp:
Some of our examples have "dev" dependencies. If possible, these should all be changed to "latest".

https://github.com/pulumi/templates/pull/44#issuecomment-441157007

PR #189, opened 2018-12-03
Use "latest" as the version in more examples

ellismg:
For a few of the examples we had harnessed, they were using older
versions of our packages. Move them to "latest" so our examples always
reflect the most recently released versions of our packages.

There are no lock file updates as part of this change as we do not
check yarn.lock's into git here.

### DIFF
--- a/aws-ts-eks/package.json
+++ b/aws-ts-eks/package.json
@@ -4,8 +4,8 @@
         "@types/node": "latest"
     },
     "dependencies": {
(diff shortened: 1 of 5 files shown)


ellismg:
> LGTM. Fixes #186?

Pushed another commit that fixed up the others (while we don't run them as part of our tests, they are in TS and so I validated that everything continued to compile).

PR #190, opened 2018-12-06
Add example for multiple AKS clusters.

clstokes:


azure-ts-aks-multicluster/Pulumi.yaml
```diff
@@ -0,0 +1,14 @@
+name: azure-ts-aks-multicluster
+runtime: nodejs
+description: Create multiple Azure Kubernetes Service (AKS) clusters in different regions and with different node counts.
+template:
+  description: Create multiple Azure Kubernetes Service (AKS) clusters in different regions and with different node counts.
```
justinvp: This line can now be deleted (we don't need to duplicate the `description` under `template` anymore).

clstokes: Removed. Thanks!

clstokes:
@justinvp is there a linter or some other way to check that I don't have extraneous properties in my `Pulumi.yaml` for next time?

That aside, does this look good enough to merge in?

azure-ts-aks-multicluster/README.md
```diff
@@ -0,0 +1,66 @@
...
+    $ pulumi stack init
+    ```
+
+2. Set the required configuration variables for this program:
+
+    ```bash
+    $ pulumi config set azure:environment public
+    $ pulumi config set password --secret [your-cluster-password-here]
+    $ ssh-keygen -t rsa -f key.rsa
+    $ pulumi config set sshPublicKey < key.rsa.pub
```
justinvp: It'd be good to mention setting the required `resourceGroupLocation` config value here as well.

azure-ts-aks-multicluster/index.ts
```diff
@@ -0,0 +1,63 @@
+// Copyright 2016-2018, Pulumi Corporation.  All rights reserved.
+
+import * as azure from "@pulumi/azure";
+import * as pulumi from "@pulumi/pulumi";
+
+// Global Config
+const config = new pulumi.Config();
+const configPassword = config.require("password");
+const configSshPublicKey = config.require("sshPublicKey");
+const configResourceGroupLocation = config.require("resourceGroupLocation");
```
justinvp: It'd be good to include this required config value in the `Pulumi.yaml`. That way:
 - If the example is used with `pulumi new` (e.g. `pulumi new https://github.com/pulumi/examples/azure-ts-aks-multicluster`), the CLI will prompt the user for the value as part of setting up the stack.
 - Or if the "Deploy with Pulumi" button in the README.MD is clicked, the config value will be included in the form that the user fills out in the browser to configure the project/stack.

Otherwise, when `pulumi new` runs the initial `up` as part of its operation, it will fail due to a missing required config value.

azure-ts-aks-multicluster/Pulumi.yaml
```diff
@@ -0,0 +1,13 @@
+name: azure-ts-aks-multicluster
+runtime: nodejs
+description: Create multiple Azure Kubernetes Service (AKS) clusters in different regions and with different node counts.
```
justinvp: Nit: the description tends to look better when displayed in the CLI without the period at the end.

PR #191, opened 2018-12-09
Add an example of "provisioning"

joeduffy:
This new example demonstrates using dynamic providers to approximate
"provisioning". It implements CopyFile and RemoteExec custom resources
that will use SCP and SSH to accomplish post-provision configuration.

The longer-term solution to this is to design and implement
https://github.com/pulumi/pulumi/issues/1691.


oliverholliday:
I've been trying to work from this example to deploy some SQL, but I get an error "Resource.isInstance is not a function". I know this isn't supported and is going to be replaced by some other method of managing resources, but do you have any ideas how I could fix that?

I'm trying to deploy an SQL schema after (azure) DB creation but short of a mechanism like for a create-hook it's not going to work.

Thanks

justinvp:
@joeduffy, do we need to keep this open or can we close?

mikhailshilkov:
@cowwoc This repo contains examples of Pulumi programs, not a reusable library. Your ability to use the code similar to these examples does not depend on whether a PR is merged on not. Basically, copy-paste the parts that you need into your program.

lukehoban:
I pushed some improvements here.

> Sometimes the SSH commands fail because the server isn't ready -- we probably need retry logic

This is fixed in latest commits.

aws-ts-ec2-provisioners/index.ts
```diff
@@ -0,0 +1,63 @@
+// Copyright 2016-2019, Pulumi Corporation.  All rights reserved.
+
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+import { CopyFile, RemoteExec } from "./provisioners";
+
+// Get the config ready to go.
+const config = new pulumi.Config();
+const publicKey = config.require("publicKey");
+const privateKey = config.require("privateKey");
```
praneetloke: Any reason not to use `requireSecret` here (and `getSecret`/`requireSecret` in the following line)?

PR #197, opened 2018-12-18
Fix location in azure-ts-aks-multicluster.

clstokes:
- per-cluster location was using global config value instead of per-cluster config
- rename location config value to be consistent with other examples

### DIFF
--- a/azure-ts-aks-multicluster/index.ts
+++ b/azure-ts-aks-multicluster/index.ts
@@ -4,7 +4,7 @@ import * as azure from "@pulumi/azure";
 import * as pulumi from "@pulumi/pulumi";
 import * as config from "./config";
 
-// Per-Cluster Config
+// Per-cluster config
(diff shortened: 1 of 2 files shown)


clstokes:
https://github.com/pulumi/examples/pull/197/files#diff-d0949064177062ff6c9b6b7ec55a4942R48 is the actual fix for the cluster location.

azure-ts-aks-multicluster/index.ts
```diff
@@ -50,6 +44,14 @@ const k8sClusters = aksClusterConfig.map((perClusterConfig, index) => {
             clientId: adApp.applicationId,
             clientSecret: adSpPassword.value,
         },
+        // Per-cluster config arguments
+        location: perClusterConfig.location,
```
clstokes: This is the actual fix for the cluster location.

PR #199, opened 2018-12-21
Create certificate if certificateArn is omitted.

clstokes:
If certificateArn is omitted from the configuration, create an ACM certificate and associated DNS records for certificate verification.

### DIFF
--- a/aws-ts-static-website/index.ts
+++ b/aws-ts-static-website/index.ts
@@ -79,11 +79,16 @@ const logsBucket = new aws.s3.Bucket("requestLogs",
 const tenMinutes = 60 * 10;
 
 let certificateArn: pulumi.Input<string> = config.certificateArn!;
+
+/**
+ * Only provision a certificate (and related resources) if a certificateArn is _not_ provided via configuration.
+ */
 if (config.certificateArn === undefined) {
-    // Per AWS, ACM certificate must be in the us-east-1 region.
+    
     const eastRegion = new aws.Provider("east", {
-        region: "us-east-1",
+        region: "us-east-1", // Per AWS, ACM certificate must be in the us-east-1 region.
     });
+
     const certificate = new aws.acm.Certificate("certificate", {
         domainName: config.targetDomain,
         validationMethod: "DNS",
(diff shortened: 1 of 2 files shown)


aws-ts-static-website/index.ts
```diff
@@ -79,6 +78,38 @@ const logsBucket = new aws.s3.Bucket("requestLogs",
...
+    });
+    const certificate = new aws.acm.Certificate("certificate", {
+        domainName: config.targetDomain,
+        validationMethod: "DNS",
+    }, { provider: eastRegion });
+
+    const domainParts = getDomainAndSubdomain(config.targetDomain);
+    const hostedZoneId = aws.route53.getZone({ name: domainParts.parentDomain }).then(zone => zone.id);
+
+    const certificateValidationDomain = new aws.route53.Record(
```
chrsmith: A comment to better explain what is going on here is probably warranted. What you are doing is creating the ACM certificate, that resource then has the `domainValidationOptions`, which you then use to create a new DNS record, which is used to verify that you own the domain. Right?

It's cool that it all "just works", but some docs explaining what's going on would be useful for the next person who looks at this code scratching their head.

aws-ts-static-website/index.ts
```diff
@@ -79,6 +78,38 @@ const logsBucket = new aws.s3.Bucket("requestLogs",
...
+        certificateArn: certificate.arn,
+        validationRecordFqdns: [certificateValidationDomain.fqdn],
+    }, { provider: eastRegion });
+
+    certificateArn = certificateValidation.certificateArn;
+}
+
 // distributionArgs configures the CloudFront distribution. Relevant documentation:
 // https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html
 // https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html
```
chrsmith: Would you mind adding a comment on top of the `aliases` attribute here too, just to make the next time this example is built upon:

```diff
 const distributionArgs: aws.cloudfront.DistributionArgs = {
     enabled: true,	     enabled: true,
+    // Alternate aliases the CloudFront distribution can be reached at, in addition to https://xxxx.cloudfront.net.
+    // Required if you want to access the distribution via config.targetDomain as well.  
     aliases: [ config.targetDomain ],
```

aws-ts-static-website/index.ts
clstokes: Yep, great point.

aws-ts-static-website/index.ts
clstokes: Added!

clstokes:
@chrsmith, added some additional comments. If you have a chance, take another quick pass and see if I missed anything, Thanks!

PR #203, opened 2019-01-07
Deny s3:ListBucket

chrsmith:
Update the static website on AWS example to deny the `s3:ListBucket` permission. We set the "public-read" ACL on the bucket, so that CloudFront can read the bucket's contents. However, just following best practices with regard to content security, we deny the `s3:ListBucket` permission to prevent unintentional disclosure of the website's contents.

See https://github.com/pulumi/get.pulumi.com/pull/33, https://github.com/pulumi/docs/pull/734


ellismg:
Do you think it is reasonable to expand the comment to say that while ListBucket is denied it is still possible for someone to guess a valid URL and in that case this infrastructure would serve the file?

PR #204, opened 2019-01-07
Update README.md

Serguzest:


aws-ts-serverless-raw/README.md
```diff
@@ -10,7 +10,7 @@ to accomplish all of the same things this higher-level package offers.
 The deployed Lambda function is a simple C# application, highlighting the ability to manage existing application code
 in a Pulumi application, even if your Pulumi code is written in a different language like JavaScript or Python.
 
-The Lambda function is a C# application using dotnet2.0 (a similar approach works for any other language supported by
+The Lambda function is a C# application using dotnet core 2.0 (a similar approach works for any other language supported by
```
lukehoban: We should probably either say either:
* `dotnetcore2.0`
* .NET Core 2.0

The former is the name of the runtime in Lambda, the latter is the product name.  I'm fine with either, and the original text here was wrong in either case :-).  Happy to change to whichever of those you'd prefer if you want to make the change.

PR #205, opened 2019-01-07
Add Webhooks handler example

chrsmith:
Add a simple Pulumi app for posting incoming Pulumi Service webhook events to Slack.

The goal is to provide a simple application developers can stand up to start writing applications that respond to Pulumi Service webhook events. I intend to link to this example from the feature's documentation when it goes live on https://pulumi.io.

The application itself isn't that interesting. It just uses the Slack SDK to post messages to a channel.  But it does provide an example of how to verify the integrity of messages received if the `pulumi-webhook-signature` header is present.

### DIFF
--- a/aws-ts-pulumi-webhooks/index.ts
+++ b/aws-ts-pulumi-webhooks/index.ts
@@ -1,6 +1,5 @@
 import * as pulumi from "@pulumi/pulumi";
 import * as cloud from "@pulumi/cloud";
-import * as cloudAws from "@pulumi/cloud-aws";
 
 import * as crypto from "crypto";
 
@@ -19,8 +18,8 @@ const stackConfig = {
 
 // Just logs information aincomming webhook request.
 async function logRequest(req: cloud.Request, _: cloud.Response, next: () => void) {
-    const webhookID = req.headers["pulumi-webhook-id"] as string;
-    const webhookKind = req.headers["pulumi-webhook-kind"] as string;
+    const webhookID = req.headers["pulumi-webhook-id"];
+    const webhookKind = req.headers["pulumi-webhook-kind"];
     console.log(`Received webhook from Pulumi ${webhookID} [${webhookKind}]`);
     next();
 }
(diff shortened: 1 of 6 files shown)


joeduffy:
So, one thing we used to do that was *very neat* with our GitHub examples, was to auto-register the webhooks as part of the creation of the Pulumi stack. Is that possible with our current APIs?

(E.g., see this and related family of GitHub APIs: https://developer.github.com/webhooks/creating/)

aws-ts-pulumi-webhooks/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    const result = crypto.timingSafeEqual(Buffer.from(webhookSig), Buffer.from(hmac));
+    if (!result) {
+        console.log(`Mismatch between expected signature and HMAC: '${webhookSig}' vs. '${hmac}'.`);
+        res.status(400).end("Unable to authenticate message: Mismatch between signature and HMAC");
+        return;
+    }
+    next();
+}
+
+const webhookHandler = new cloudAws.HttpEndpoint("pulumi-webhook-handler");
```
lukehoban: May as well just use `cloud` here since you ended up needing to import it - you can then remove the import of `cloudAws`.

lukehoban:
> So, one thing we used to do that was very neat with our GitHub examples, was to auto-register the webhooks as part of the creation of the Pulumi stack. Is that possible with our current APIs?

We do have the APIs, though like the rest of the Pulumi Service REST API, they are not (yet) documented.

We also don't (yet) have a Pulumi Pulumi Provider which would allow automating the use of those REST APIs via a Pulumi program, though there is nothing technically preventing that from being done.

While it would be possible and nice to add those, I do think a simple example like this is a very nice starting point.

joeduffy:
> While it would be possible and nice to add those, I do think a simple example like this is a very nice starting point.

I's also unfortunate for us not to use our platform to make this example really shine, missing an opportunity to remove manual steps for installing and configuring your cloud software.

joeduffy:
Definitely not blocking this going in, just submitting a request for a nice add on example once we feel our API is there. Also mentioning it so that we collectively (ahem, Chris) think about this as an API that people will want to automate (in other words, it doesn't exist solely to serve our console).

aws-ts-pulumi-webhooks/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+webhookHandler.get("/", async (_, res) => {
+    res.status(200).end("🍹 Pulumi Webhook Responder🍹\n");
+});
+
+webhookHandler.post("/", logRequest, authenticateRequest, async (req, res) => {
+    const webhookKind = req.headers["pulumi-webhook-kind"] as string;
+    const payload = <string>req.body.toString();
```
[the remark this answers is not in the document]
chrsmith: The indexer on `req.headers` returns `string | string[]`. So in at least one case the `as string` cast is necessary. But I'll remove the cases where it doesn't matter.

aws-ts-pulumi-webhooks/index.ts
chrsmith: SGTM

PR #209, opened 2019-01-09
Remove bucket policy.

clstokes:
Reverts pulumi/examples#203

While the bucket policy is a good practice, it can cause workflow issues for users as it causes `pulumi refresh` to fail. Reverting this change until we can put together a solution that supports preventing listing the bucket anonymously while supporting the pulumi workflow.

### DIFF
--- a/aws-ts-static-website/index.ts
+++ b/aws-ts-static-website/index.ts
@@ -31,25 +31,6 @@ const contentBucket = new aws.s3.Bucket("contentBucket",
         }
     });
 
-// contentBucket needs to have the "public-read" ACL so its contents can be ready by CloudFront and
-// served. But we deny the s3:ListBucket permission to prevent unintended disclosure of the bucket's
-// contents. If you know the Bucket object path, it is still available for anonymous access however.
-const denyListPolicyState: aws.s3.BucketPolicyArgs = {
-    bucket: contentBucket.bucket,
-    policy: contentBucket.arn.apply(arn => JSON.stringify({
-        Version: "2008-10-17",
-        Statement: [
-            {
-                Effect: "Deny",
-                Principal: "*",
-                Action: "s3:ListBucket",
-                Resource: arn,
-            },
-        ],
-    })),
-};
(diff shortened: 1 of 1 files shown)


joeduffy:
Interesting! I'm just curious, what workflow issues does this cause, and why does `refresh` fail?

clstokes:
`refresh` failing is the workflow issue. The refresh fails because it attempts to list the bucket and the list fails due to the bucket policy. The bucket policy essentially blocks everyone (anonymous, authenticated users, and the root user) from listing the bucket.

clstokes:
The refresh error is at https://gist.github.com/clstokes/b0218f967d9d2e840781d353475323c2#file-gistfile1-txt-L18-L21.

joeduffy:
Ah, got it. Amusing ... (in a not-so-amusing kind of way.)

lukehoban:
I can't actually tell why that error should happen.  The provider doesn't call `ListBuckets` anywhere.  I believe that some calls to `GetObject` will fail with `403` instead of `404` when this permission is denied, but not sure why the provider would be requesting buckets that don't exist during refresh either.  Possibly worth opening a separate issue to track this, as it will bite folks (including us) trying to follow best practices.

clstokes:
It's the HeadBucket call that's failing - https://github.com/terraform-providers/terraform-provider-aws/blob/0724f51c3240a7c039c144e44650cab91fda50a2/aws/resource_aws_s3_bucket.go#L687.

Full Terraform output and `s3api head-bucket` call failing too - https://gist.github.com/clstokes/1671ee3e352c8cd659e81ff32788a7d4#file-_tf-log-L27-L29.

I'll open an issue in the terraform-provider-aws repo.

lukehoban:
Indeed - and looks like this is explicitly documented by AWS (though I can't tell the rationale and it seems a bit awkward):

https://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html#using-with-s3-actions-related-to-buckets

PR #211, opened 2019-01-10
Add Lambda@Edge example to aws-ts-static-website.

clstokes:


aws-ts-static-website/index.ts
```diff
@@ -145,14 +126,73 @@ if (config.certificateArn === undefined) {
     certificateArn = certificateValidation.certificateArn;
 }
 
+/**
+ * Only provision a Lambda@Edge fuction (and related resources) if languageRedirect is 'true' from configuration.
+ * Once provisioned test with the following curl commands
+ * - curl -L -H 'accept-language: en' $(pulumi stack output targetDomainEndpoint)
+ * - curl -L -H 'accept-language: es' $(pulumi stack output targetDomainEndpoint)
+ */
+let languageCacheBehavior = [];
```
lukehoban: Perhaps name this `cacheBehaviours`?

aws-ts-static-website/index.ts
```diff
@@ -145,14 +126,73 @@ if (config.certificateArn === undefined) {
...
+
+    const languageRedirectLambda = new aws.lambda.Function("language-redirect", {
+        code: new pulumi.asset.AssetArchive({
+            "index.js": new pulumi.asset.FileAsset("./src/lambda-language-redirect.js"),
+        }),
+        role: languageRedirectRole.arn,
+        handler: "index.handler",
+        publish: true,
+        runtime: aws.lambda.NodeJS8d10Runtime,
+    }, { provider: eastRegion });
```
lukehoban: Is this not supported (yet) with `aws.lambda.CallbackFunction`?  That would in principle allow you to write the handler in-line here - which could be quite slick for this sort of use case (very small lambdas generally for Lambda@Edge, and very much part of the infrastructure behavior). 

Cc @cyrusnajmabadi.

CyrusNajmabadi: agreed. this looks like this should be doable using 'aws.lambda.CallbackFunction'.  Ideally it is self explanatory to do so.  If you try it out and it doesn't work, please feel free to ping me on slack to help you through it.

aws-ts-static-website/src/lambda-language-redirect.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
```
CyrusNajmabadi: as per above, we could likely just have this be inlined as a javascript function whne making the AWS lambda.  Also, it *should* be possible to just make this 'async' instead of the 'callback' form.  But you could definitely just start with directly inlining this first, then trying out the async option afterwards if that's not the case.

aws-ts-static-website/index.ts
```diff
@@ -191,6 +231,8 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
         maxTtl: tenMinutes,
     },
 
+    orderedCacheBehaviors: cacheBehaviours,
```
jenstroeger: @clstokes, I think we talked about this offline but when I saw the code then my question was why the `cacheBehaviors` is attached to `orderedCacheBehaviors` here instead of added to the defaults. Perhaps mention your reasoning in a comment?

aws-ts-static-website/src/lambda-language-redirect.js
```diff
@@ -0,0 +1,32 @@
...
+    const languageHeader = request.headers['accept-language'];
+
+    // shortcut invalid requests
+    if (request.uri != '/' || typeof languageHeader === 'undefined') {
+        callback(null, request);
+        return;
+    }
+
+    const clientLanguages = languageHeader[0].value;
+    console.log('accept-language header from client:', clientLanguages);
```
jenstroeger: I still haven’t found out where these logs go! The [documentation](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-edge-testing-debugging.html) says that this will be logged through Cloudwatch and in the nearest region, but that didn’t work for me. Would the logging require additional AWS infrastructure in the Pūlumi program?

aws-ts-static-website/www/es/index.html
```diff
@@ -0,0 +1,18 @@
...
+      body {
+        background-color: lightblue;
+      }
+  </style>
+</head>
+
+<body>
+    <h1>¡Hola, mundo!</h1>
+    <p>Hecho con ♥ usando <a href="https://pulumi.io">Pulumi</a>.</p>
+</body>
```
jenstroeger: Missing `</html>` tag?

mikhailshilkov:
@clstokes Do you remember why this was never finished? Worth adding last bits and merge?

clstokes:
@mikhailshilkov I didn't get back to this to implement Cyrus's suggestions around async and CallbackFunction. If you're able to implement that, then please feel free to go ahead. Otherwise I will get back to this eventually.

mikhailshilkov:
@clstokes I refactored the Lambda, also merged master, fixed a couple of other comments and one compile error. Please have a look and test it - I only tested the Lambda itself, my AWS account can't make CloudFront and I don't have a domain to try.

aws-ts-static-website/src/lambda-language-redirect.js
mikhailshilkov: An extra policy is required to allow the lambda writing the logs, see my latest commit

aws-ts-static-website/src/lambda-language-redirect.js
mikhailshilkov: Done

aws-ts-static-website/index.ts
mikhailshilkov: Done

stack72:
Closing this PR our as stale - there has been no activity in almost 2 years. Please rebase and reopen if you think this is a valuable addition

PR #214, opened 2019-01-30
Add some example webhook formatting

cnunciato:
This change adds a utility function to `aws-ts-pulumi-webhooks`  in order to provide users with a few examples to work by in formatting their own messages. E.g.,

image

Signed-off-by: Christian Nunciato <c@nunciato.org>

### DIFF
--- a/aws-ts-pulumi-webhooks/index.ts
+++ b/aws-ts-pulumi-webhooks/index.ts
@@ -58,10 +58,10 @@ webhookHandler.post("/", logRequest, authenticateRequest, async (req, res) => {
     const webhookKind = req.headers["pulumi-webhook-kind"].toString();
     const payload = <string>req.body.toString();
     const prettyPrintedPayload = JSON.stringify(JSON.parse(payload), null, 2);
-    const fallbackText = `Pulumi Service Webhook (\`${webhookKind}\`)\n` + "```\n" + prettyPrintedPayload + "```\n";
 
     const client = new slack.WebClient(stackConfig.slackToken);
 
+    const fallbackText = `Pulumi Service Webhook (\`${webhookKind}\`)\n` + "```\n" + prettyPrintedPayload + "```\n";
     const messageArgs: slack.ChatPostMessageArguments = {
         channel: stackConfig.slackChannel,
         text: fallbackText,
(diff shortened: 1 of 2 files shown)


aws-ts-pulumi-webhooks/index.ts
```diff
@@ -53,19 +55,23 @@ webhookHandler.get("/", async (_, res) => {
 });
 
 webhookHandler.post("/", logRequest, authenticateRequest, async (req, res) => {
-    const webhookKind = req.headers["pulumi-webhook-kind"];
+    const webhookKind = req.headers["pulumi-webhook-kind"].toString();
     const payload = <string>req.body.toString();
     const prettyPrintedPayload = JSON.stringify(JSON.parse(payload), null, 2);
+    const fallbackText = `Pulumi Service Webhook (\`${webhookKind}\`)\n` + "```\n" + prettyPrintedPayload + "```\n";
```
chrsmith: Nit: Since this is only used inside of `messageArgs`, can we put its declaration to right above it?

cnunciato: > And my understanding is that TypeScript will then figure out the right overload to call based on that string's value.

Indeed, I was not aware of this fanciness! After a good bit of fiddling, though, I couldn't quite bring it together, so I've pulled the formatters into functions as you suggested, and that does feel tidier, if not quite ✨.

cnunciato:
> And my understanding is that TypeScript will then figure out the right overload to call based on that string's value.

Indeed, I was not aware of this fanciness! After a good bit of fiddling, though, I couldn't quite bring it together, so I've pulled the formatters into functions as you suggested, and that does feel tidier, if not quite ✨.

aws-ts-pulumi-webhooks/index.ts
```diff
@@ -53,19 +55,23 @@ webhookHandler.get("/", async (_, res) => {
 });
 
 webhookHandler.post("/", logRequest, authenticateRequest, async (req, res) => {
-    const webhookKind = req.headers["pulumi-webhook-kind"];
+    const webhookKind = req.headers["pulumi-webhook-kind"].toString();
```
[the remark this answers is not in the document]
cnunciato: Oh, ok. I was assuming that since these requests originate from the Pulumi service, we'd be able to rely on the presence of this header, but if that's an unsafe assumption, I can add a little error handling here as well.

PR #229, opened 2019-02-09
GKE & EKS: add clarifying info based on TGIK8s feedback

metral:


joeduffy:
The other thing we should do is probably remove the paragraph about changing configuration, or at least move it elsewhere (to before the initial `pulumi up`). Most of the time, someone doesn't want to tear down and recreate the entire cluster when changing those parameters. To do an in place upgrade would require that the cluster configuration use node pools, probably too complex for this example.

metral:
> The other thing we should do is probably remove the paragraph about changing configuration, or at least move it elsewhere (to before the initial `pulumi up`). Most of the time, someone doesn't want to tear down and recreate the entire cluster when changing those parameters. To do an in place upgrade would require that the cluster configuration use node pools, probably too complex for this example.

I call this out in https://github.com/pulumi/examples/pull/229/files#diff-b7ae66bbc6b58d9e4a56d7eeb299ff14R66 - should we just bubble it up to before the first `pulumi up`?

metral:
@joeduffy I've removed the config modification + nodepool section as it was out of scope for this tutorial.

Reminder, this is in the examples repo, so these docs are a bit bare compared to the more complete, corresponding tutorial in `pulumi/docs`. I'm tracking the same feedback changes for the tutorial in https://github.com/pulumi/docs/pull/869 if that helps provide more clarity on seeing what changes made it in / still need work.

joeduffy:
> I've removed the config modification + nodepool section as it was out of scope for this tutorial.

That feels like the right call to me. I'd love to have this in the examples, somewhere, as it's way above my pay-grade but a very obvious question to ask ("How do I scale my cluster up without downtime?")

metral:
> > I've removed the config modification + nodepool section as it was out of scope for this tutorial.
> 
> That feels like the right call to me. I'd love to have this in the examples, somewhere, as it's way above my pay-grade but a very obvious question to ask ("How do I scale my cluster up without downtime?")

Ya, the upgrade story is straight forward on GKE, but not so much on EKS, so that'll require more work in general outside of this PR, but agreed that we should aim for that in the examples when we have a better handle on it.

How does the overall feedback incorporated look for this and its more user-facing complement (https://github.com/pulumi/docs/pull/869). Any feedback? Do the changes address the points made?

aws-ts-eks-hello-world/README.md
```diff
@@ -205,23 +225,21 @@ After cloning this repo, from this working directory, run these commands:
...
+    > point to the new resource, before the old resource is deleted.
+    > This is generally quite useful.
+    >
+    > ```
+    > deploymentName : "helloworld-58jkmc7c"
+    > ...
+    > namespaceName  : "helloworld-xaldhgca"
+    > serviceHostname: "a71f5ab3f2a6e11e3ac39200f4a9ad5d-1297981966.us-west-2.elb.amazonaws.com"
+    > serviceName    : "helloworld-3fc2uhh7"
+    > ```
```
lukehoban: I think we can replace all this with just:

```
**Note:** Pulumi auto-generates a suffix for all objects.  For more details on why see [documentation on resource names](https://pulumi.io/reference/programming-model.html#names).
```

That linked page is being deployed right now: https://github.com/pulumi/docs/pull/871.

gcp-ts-gke-hello-world/README.md
```diff
@@ -32,33 +36,37 @@ After cloning this repo, from this working directory, run these commands:
...
 
+    This installs the dependent packages needed for our Pulumi program.
+
+    Because we use TypeScript to author our Pulumi program, we must install
+    the packages needed for Pulumi's *language host,* which coordinates with
+    the [Pulumi service](https://app.pulumi.com) to locally observe and execute the program on your
+    machine.
+
+    Learn [how Pulumi works](https://pulumi.io/reference/how.html) if you're
+    interested in peeking under the hood.
```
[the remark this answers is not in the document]
metral: Sounds good. The added detail was w.r.t him asking why he needed node in the first place, so I was trying to scratch that itch. But I agree, your tip to link to the docs is cleaner.

aws-ts-eks-hello-world/README.md
metral: 👍

metral:
@joeduffy @lukehoban I believe I've incorporated all feedback from TGIK and listed here. PTAL

PR #234, opened 2019-02-15
Add an example of deploying a Nginx server in GCP

zzxwill:
In a GCP instance, deploy a Nginx server by `gcp-py` environment.

### DIFF
--- a/gcp-py-instance-nginx/README.md
+++ b/gcp-py-instance-nginx/README.md
@@ -1,0 +1,237 @@
+[![Deploy](https://get.pulumi.com/new/button.svg)](https://app.pulumi.com/new)
+
+# Pulumi Nginx Server in a instance (GCP)
+
+Starting point for building the Pulumi nginx server sample in Google Cloud Platform.
+
+## Running the App
+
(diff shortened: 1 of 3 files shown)


gcp-py-instance-nginx/README.md
```diff
@@ -0,0 +1,237 @@
...
+    ```
+
+2.  Configure the project:
+
+    ```
+    $ export GOOGLE_PROJECT=cncf-230209; export GOOGLE_REGION=asia-east1; export GOOGLE_ZONE=asia-east1-a;
+    $ export GOOGLE_CREDENTIALS=YOURGCPCREDENTIALS
+    ```
+
+3.  Run `pulumi up` to preview and deploy changes:
```
swgillespie: Before we run `pulumi up` I think we need to configure the Python project:

```
virtualenv -p python3 venv
source venv/bin/activate
pip install pulumi_gcp
```

Can you add this between steps 2 and 3?

zzxwill: Done. Thanks for the suggestion:)

PR #237, opened 2019-02-21
azure aks example

4c74356b41:
example corresponding to my doc page.


4c74356b41:
@swgillespie

azure-py-aks/__main__.py
```diff
@@ -0,0 +1,125 @@
...
+from pulumi_azure.core import ResourceGroup
+from pulumi_azure.role import Assignment
+from pulumi_azure.ad import Application, ServicePrincipal, ServicePrincipalPassword
+from pulumi_azure.containerservice import KubernetesCluster, Registry
+from pulumi_azure.network import VirtualNetwork, Subnet
+
+config = pulumi.Config('azure-py-kubernetes')
+PREFIX = config.get('prefix') or 'replaceme'
+PASSWORD = config.get('password') or 'DEf34tg!rg35y#F23'
+SSHKEY = config.get('sshkey') or 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCxinIAIDDCradZPAgX5GzBLv00u4rigOLUbU00E44FrfMTqu5wXiejJ4ycSb1bI+//ZNgaB2UYRbPL7A9OUKY+K4sX5O84Q6DPMjo/90IANHVTLf3xTaSc7hpvXOtIjJTJeiamxClgnTAcR55RV/j9/Wptxa8GGcRmRCcSmJUkx5AZTFI+s8aF0W3aeHHRw7TxNKBuwrX7FDcHyGKvdkFg4OP863Xe5hp5ql1C3XibmCOp1CMPIU2hCmGOy1LGbOf/Pa+QKAdtUSrPNK/jBWvPWo0k02Ii0JtMAdlpVqnJc3czNIp5gEqZCRCGEdkb/kZnJiMRZhmLBYnC8tiMxvZj core@k8s'
```
hausdorff: You probably don't want your public SSH key in this PR. :) For all of these config valiables, I think we should just use `config.required(...)` instead.

azure-py-aks/__main__.py
```diff
@@ -0,0 +1,125 @@
...
+LOCATION = config.get('location') or 'westeurope'
```
hausdorff: I think this is not a real Azure region. Can we default to `East US`, or `config.require`?

4c74356b41: why not, it is a real azure region ;)

4c74356b41: i can switch to `eastus` if you want

azure-py-aks/__main__.py
4c74356b41: its a random one I generated for folks so they dont have to do it. the idea here (because its a learning experience) that it will just work. I can make them required

azure-py-aks/README.md
```diff
@@ -0,0 +1,67 @@
...
+5. After 10-15 minutes, your cluster will be ready, and the kubeconfig YAML you'll use to connect to the cluster will be available as an output. You can save this kubeconfig to a file like so:
+
+    ```bash
+    $ pulumi stack output kubeconfig >kubeconfig.yaml
+    ```
+
+    Once you have this file in hand, you can interact with your new cluster as usual via `kubectl`:
+
+    ```bash
+    $ KUBECONFIG=./kubeconfig.json kubectl get nodes
```
hausdorff: Also this should be `./kubeconig.yaml` right?

azure-py-aks/__main__.py
hausdorff: I actually would prefer it be in the US since most of our users are probably in the US, but it's surprising to me that `East US` and `eastus` are both valid regions. But, I am not well versed in Azure.

azure-py-aks/__main__.py
hausdorff: Makes sense, if you do make it required I think you can probably tell people to do something like `pulumi config set sshkey < ~/.ssh/id_rsa.pub` and it should be about as easy.

azure-py-aks/README.md
```diff
@@ -0,0 +1,67 @@
...
+    $ pulumi stack init
+    ```
+
+3. Set the configuration variables for this program, they have default values, so you dont have to do that:
+
+    ```bash
+    $ pulumi config set prefix all_resources_will_be_prefixed_with_this_value
+    $ pulumi config set password service_principal_password
+    $ pulumi config set sshkey your_public_ssh_key
+    $ pulumi config set location any_valid_azure_location_for_aks
```
hausdorff: While you're making changes, if you omit the `$` at the beginning of these code blocks, it makes it much easier to copy and paste blocks of code. We'd still have to change the values of the config variables, but at least you can copy+paste once instead of 4 times.

4c74356b41: thats quite the opposite of what I was asked in the other PR. I submitted it without `$` and was asked to add those back....

hausdorff: Technically our style is to add `$` when there is output and omit it when there is not. If that's not what you were told, then that was wrong. :) This is is meant to conform with how ~all markdown linters.

hausdorff: But, it's also not a point worth dying over, which is why I bring it up in an offhand way.

4c74356b41: https://github.com/pulumi/docs/pull/895

hausdorff: cc @justinvp. See above.

hausdorff: Let's not block on this since we're already inconsistent anyway.

4c74356b41: ok, i think I've resolved the other ones.

PR #257, opened 2019-03-11
Wait for service to come up before tearing it down

CyrusNajmabadi:


aws-js-containers/index.js
```diff
@@ -17,5 +16,12 @@ let service = new awsx.ecs.FargateService("nginx", {
     },
 });
 
-// export just the hostname property of the container frontend
-exports.hostname = pulumi.interpolate `http://${listener.endpoint.hostname}`;
+// expose some APIs meant for testing purposes.
+let api = new awsx.apigateway.API("containers", {
```
lukehoban: Why is this needed?

Unfortunately - I don't think we should put anything in the examples that is overtly for testing purposes.  The examples really do need to stay focused on being user-facing examples.  If we can find ways to make them more testable while still being great purely as examples - we should do that.

CyrusNajmabadi: gotcha... the main issue here is one of timing.  it looks like there's an issue with terraform where they can end up doing things badly if destroy happens quickly after create: https://github.com/hashicorp/terraform/issues/2902

They identify a potential problem being that the service should not be destroyed here before rolepolicies it depends on are destroyed.  However, in our case, becaues we're using awsvpc we don't pass along roles/policies to our service.  

This was a attempt at a cludgy workaround by at least having some validation wait until the service was up and running properly before tearing it down.  I'm not quite sure the best way to actually handle this.  Note: i don't really mind having the API here.  It doesn't seem unreasonable given that the example if launching an nginx instance and pointing to some html.

CyrusNajmabadi:
I'm going to merge in for now to help with test stability.  I'm also opening a tracking issue of a better way to deal with this general issue.

PR #275, opened 2019-04-05
Fix aws-py-s3-folder test

swgillespie:
Simple fix, the stack output is called `website_url` and not `websiteUrl`.

Also fixed the test runner to run tests by their base directory name instead of their file path, which makes it much easier to run individual tests locally using `go test`.


swgillespie:
Going to go ahead and merge this to unbreak the CI.

PR #281, opened 2019-04-17
Add Null Assertion operator to pathParameters!["route"] Access

jestrjk:
This is a small change from the previous comments on this issue where as the event variable wasn't the problem, being passed in from the event Handler calling mechanism. It was accessing an untyped(?) event object property `pathParameters` which was causing the TS compile error.

Our ts compiler is slightly different versions mine being:
```
$tsc --version
Version 3.5.0-dev.20190409
```


ellismg:
>  It was accessing an untyped(?) event object property pathParameters which was causing the TS compile error.

It actually looks like the type just includes null as a possibility.  Was surprised this ever worked!  Thank you for the fix!

PR #283, opened 2019-04-18
Fix TypeScript compilation error in aws-ts-apigateway

ellismg:
We know the pathParameters array will not be null, because we capture
parameters in our route.


ellismg:
Our nightly CI Job hit this: https://travis-ci.com/pulumi/examples/jobs/193926229

CyrusNajmabadi:
INteresting. i wonder if TS got stricter.  This was fine when originally written...

PR #285, opened 2019-04-25
Refactor azure-ts-functions to use azure-serverless package

mikhailshilkov:
Stop using hand-made `HttpFunction`, use one from `@pulumi/azure-serverless` instead


mikhailshilkov:
Not applicable anymore, will open a new one soon

PR #291, opened 2019-05-01
App Service + DevOps sample

mikhailshilkov:
A Todo List application from Azure Samples GitHub: [.NET Core MVC sample for Azure App Service](https://github.com/azure-samples/dotnetcore-sqldb-tutorial), a web app built with ASP.NET Core, Entity Framework Core and a SQL database. 

Provisions Azure SQL Database and Azure Application Insights to be used in combination with App Service. Defines an Azure DevOps pipeline to deploy in CI/CD environment.


lukehoban:
cc @praneetloke as well to take a look at the DevOps integration, which we may want to align with the new work we're landing on simpler DevOps Pipelines integration right now.

mikhailshilkov:
@lukehoban A dry run of this sample deployment took 3 minutes 18 seconds.

azure-ts-appservice-devops/azure-pipelines.yml
```diff
@@ -0,0 +1,26 @@
...
+    vmImage: 'ubuntu-16.04'
+  steps:
+  - task: DotNetCoreCLI@2
+    inputs:
+      command: 'publish'
+      projects: 'src'
+      zipAfterPublish: false
+      modifyOutputPath: false
+    displayName: 'Build and publish ASP.NET Core app'
+  - script: |
```
praneetloke: Yeah it would be great to use the task extension available in public preview here: https://marketplace.visualstudio.com/items?itemName=pulumi.build-and-release-task.

You could then convert this `script` task to ditch the bash scripts:

```yaml
- task: Pulumi@0
    inputs:
      azureSubscription: "Service Connection to your Azure Subscription"
      command: "up"
      args: "--yes"
      cwd: "pulumi-stack/"
      stack: "dev"

# At this point the pulumi tool should be in the path of the agent. So you should be able to use it in a subsequent step, but within the same job.
- script: |
    # Set config properties
     pulumi config set --secret sqlUsername ${SQL_USERNAME}
     pulumi config set --secret sqlPassword ${SQL_PASSWORD}

# Or, you can also use the Pulumi task once again. This may be a little verbose than just running an inline script like above if you are wanting to run `config set...` commands.
- Pulumi@0
   inputs:
     azureSubscription: "Service Connection to your Azure Subscription"
     command: "config"
     args: "set --secret sqlUserName $(sql.username)"
     cwd: "pulumi-stack/"
     stack: "dev"

- Pulumi@0
   inputs:
     azureSubscription: "Service Connection to your Azure Subscription"
     command: "config"
     args: "set --secret sqlPassword $(sql.password)"
     cwd: "pulumi-stack/"
     stack: "dev"
```

@mikhailshilkov would you mind installing the extension to your organization and giving it a try with your example here? I would love your feedback on it! Afterwards, it would be great to include an example of the pipeline config that shows how to use the task extension as well in addition to the current one that you have here.

I think we should still include merge this example in its current form, since not everyone may have access to installing extensions into their organization, and as such I don't want folks trying the example to be blocked by not being able to install the extension.

mikhailshilkov: Done! Thanks for the task, it simplifies the setup a lot!

azure-ts-appservice-devops/azure-pipelines.yml
```diff
@@ -10,17 +10,19 @@ jobs:
...
-  - script: |
-      chmod +x infra/scripts/*.sh
-      ./infra/scripts/setup.sh
-      ./infra/scripts/run-pulumi.sh
+  - task: Npm@1
+    inputs:
+      command: 'install'
+      workingDir: 'infra'  
+    displayName: 'NPM install packages'
+  - task: Pulumi@0
```
praneetloke: Great to see this here already! You rock, @mikhailshilkov!

azure-ts-appservice-devops/azure-pipelines.yml
```diff
@@ -10,17 +10,19 @@ jobs:
...
     env:
-      PULUMI_ACCESS_TOKEN: $(pulumi.access.token)
-      ARM_CLIENT_ID: $(arm.client.id)
-      ARM_TENANT_ID: $(arm.tenant.id)
-      ARM_CLIENT_SECRET: $(arm.client.secret)
-      ARM_SUBSCRIPTION_ID: $(arm.subscription.id)
-      SQL_USERNAME: ${sql.username}
-      SQL_PASSWORD: ${sql.password}
\ No newline at end of file
+      PULUMI_ACCESS_TOKEN: $(pulumi.access.token)
```
praneetloke: Also, I realize our current Azure DevOps docs shows that you need to map the `PULUMI_ACCESS_TOKEN` env var, but since you are using the Pulumi task here, you no longer need to do that. As long as you define the `pulumi.access.token` build variable in your pipeline build, the task should be able to automatically map that build var to the env var. This isn't obvious in any of the docs for the Pulumi task, which I will fix today.

mikhailshilkov: That's great! Tested and removed.

azure-ts-appservice-devops/README.md
```diff
@@ -0,0 +1,88 @@
...
+    <head>
+        <meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>Index - My ToDoList App</title>
+    ...
+    ```
+
+## Integrating with Azure DevOps
+
+`azure-pipeline.yml` in the root folder of this example shows a configuration for Azure DevOps using [Pulumi task](https://marketplace.visualstudio.com/items?itemName=pulumi.build-and-release-task).
```
praneetloke: Do you mind adding a note here about the `pulumi.access.token` build variable and that one can get a value for it from https://app.pulumi.com/account/tokens? It isn't clear from just looking at these build configuration files where one might get that. And looking at our pulumi.io site, I can't find a page that shows that either, though, it is referenced by the `pulumi login` command. I'll work with someone on the team to fix that.

praneetloke:
This looks good to me. Just one last comment about adding a note about `pulumi.access.token` in the readme file.

azure-ts-appservice-devops/README.md
mikhailshilkov: Added a line to README, please check if you think it's descriptive enough.

praneetloke: Looks good to me, @mikhailshilkov. Thank you!

lukehoban:
This looks great!  Going to go ahead and merge in now.  If there are further improvements needed, feel free to open a follow-up PR.  Thanks!

PR #292, opened 2019-05-01
Mutli-region Functions and Cosmos sample

mikhailshilkov:
Multi-region deployment of Azure Functions and Cosmos DB with Traffic Manager

### DIFF
--- a/azure-ts-serverless-url-shortener-global/index.ts
+++ b/azure-ts-serverless-url-shortener-global/index.ts
@@ -1,0 +1,120 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as azure from "@pulumi/azure";
+import { Container } from "@azure/cosmos";
+import { getContainer } from "./cosmosclient";
+
+// Read a list of target locations from the config file:
+// Expecting a comma-separated list, e.g., "westus,eastus,westeurope"
+const config = new pulumi.Config();
+const locations = config.require("locations").split(',');
+// The first location is considered primary
+const primaryLocation = locations[0];
+
+let resourceGroup = new azure.core.ResourceGroup("urlshort-rg", {
+    location: primaryLocation,
+});
+
+// Cosmos DB with a single write region (primary location) and multiple read replicas
+let cosmosdb = new azure.cosmosdb.Account("url-cosmos", {
+    resourceGroupName: resourceGroup.name,
+    location: primaryLocation,
+    geoLocations: locations.map((location, priority) => ({
+        location: location,
+        failoverPriority: priority,
+    })),
+    offerType: "Standard",
+    consistencyPolicy: {
+        consistencyLevel: "Session",
+        maxIntervalInSeconds: 5,
+        maxStalenessPrefix: 100,
+    },
+});
+
+// Traffic Manager as a global HTTP endpoint
+const profile = new azure.trafficmanager.Profile("urlshort-tm", {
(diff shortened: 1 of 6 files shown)


azure-ts-serverless-url-shortener-global/index.ts
```diff
@@ -0,0 +1,120 @@
...
+        resourceGroup,
+        location,
+        route: "{key}",
+        callbackFactory: () => {
+            const endpoint = cosmosdb.endpoint.get();
+            const masterKey = cosmosdb.primaryMasterKey.get();
+
+            let container: Container;
+            return async (_, request: azure.appservice.HttpRequest) => {
+                container = container || await getContainer(endpoint, masterKey, location);
```
CyrusNajmabadi: is this line expected to be expensive?  if so, callbackFactory is appropriate to use here.  If it's not expensive, i would just use `callback: async () ... ` and just compute this on demand here.

azure-ts-serverless-url-shortener-global/index.ts
```diff
@@ -0,0 +1,120 @@
...
+import { getContainer } from "./cosmosclient";
+
+// Read a list of target locations from the config file:
+// Expecting a comma-separated list, e.g., "westus,eastus,westeurope"
+const config = new pulumi.Config();
+const locations = config.require("locations").split(',');
+// The first location is considered primary
+const primaryLocation = locations[0];
+
+let resourceGroup = new azure.core.ResourceGroup("urlshort-rg", {
```
CyrusNajmabadi: is `-rg` a common azure naming strategy?  if so, this is fine.  But otherwise, i think we'd generally prefer the names not contain these sorts of hungarian notations.

azure-ts-serverless-url-shortener-global/index.ts
```diff
@@ -0,0 +1,120 @@
...
+});
+
+// Cosmos DB with a single write region (primary location) and multiple read replicas
+let cosmosdb = new azure.cosmosdb.Account("url-cosmos", {
+    resourceGroupName: resourceGroup.name,
+    location: primaryLocation,
+    geoLocations: locations.map((location, priority) => ({
+        location: location,
+        failoverPriority: priority,
+    })),
```
CyrusNajmabadi: could be simplified as:  `locations.map((location, failoverPriority) => ({ location, failoverPriority })`

azure-ts-serverless-url-shortener-global/index.ts
mikhailshilkov: It is sort of expensive because it checks if the database and the container exist and creates them otherwise. Ideally, I would put this initialization outside of the callback, but `await` is not allowed there.
For this example, we can change back to `callback` to simplify, but in my blog I'm making a case for fast response time, and those call screw the latency.

mikhailshilkov: Well, ideally ideally, the database and container would be created by ARM/Pulumi, but they aren't managed by ARM (yet).

azure-ts-serverless-url-shortener-global/index.ts
mikhailshilkov: No, it's not common. I'll try to find better names.

lukehoban:
Looks great!  Merging now as this is in good state . If any further updates are needed feel free to send follow-up PRs.

PR #296, opened 2019-05-06
Upgrade `pip` inside container

ellismg:
We are seeing some issues in CI where `awscli` fails to install with
errors deep in the bowels of pip. I'm unable to reproduce this
locally, but I suspect that there's an network issue in Travis that is
causing us to need to retry downloads, which is hitting some bad path
inside the very old version of `pip` we have

Let's try upgrading it to the latest version (19.X vs 8.X!) and see if
it helps.


ellismg:
Suspect this was triggered by: https://www.traviscistatus.com/incidents/kyf149kl6bvp, since from the traces it very much looked like we were hitting networking errors.  Either the incident has been better mitigated or this change actually provides help during the docker build for pip installs when they need to retry.

ellismg:
Gonna go ahead and merge this, @lukehoban to try to get the builds green again.  Let me know if you have any feedback here.

PR #302, opened 2019-05-20
Azure AppService + Docker

mikhailshilkov:
Add an Azure AppService/Linux/Docker example as discussed in https://github.com/pulumi/pulumi-azure/issues/228 and on Slack


mikhailshilkov:
tagging @lukehoban

azure-ts-appservice-docker/README.md
```diff
@@ -0,0 +1,69 @@
...
+1.  Restore NPM dependencies:
+
+    ```
+    $ npm install
+    ```
+
+1. Clone a sample Docker image from GitHub. If you run on Windows, be sure to keep Linux line endings:
+
+    ```
+    git clone -c core.autocrlf=false https://github.com/Azure-Samples/docker-django-webapp-linux
```
lukehoban: Perhaps instead of asking users to clone this, we can just include a simple Dockerfile (and app if needed) inside this repo?  We do that in several other examples.

mikhailshilkov: I included the whole sample app as-is. Let me know if it's too big and I'll search for a simpler one.

stack72:
@lukehoban I am happy with this if you are - this will also close out https://github.com/pulumi/pulumi-azure/issues/228 AFAICT - right @mikhailshilkov ?

mikhailshilkov:
@stack72 Yes. That issue can be closed as the support is there, and this example shows how to get started.

stack72:
I was able to test this and it's working as expected - thanks so much @mikhailshilkov

PR #304, opened 2019-05-24
Managing Secrets and Secure Access in Azure Applications

mikhailshilkov:
The application consists of several parts:

- An ASP.NET Application which reads data from a SQL Database and from a file in Blob Storage
- App Service which host the application. The application binaries are placed in Blob Storage, with Blob Url placed as a secret in Azure Key Vault
- App Service has a Managed Identity enabled
- The identify is granted access to the SQL Server, Blob Storage, and Key Vault
- No secret information is placed in App Service configuration: all access rights are derived from Active Directory


azure-ts-msi-keyvault-rbac/index.ts
```diff
@@ -0,0 +1,170 @@
...
+    storageAccountName: storageAccount.name,
+    containerAccessType: "private",
+});
+
+// Azure SQL Server that we want to access from the application
+const sqlServer = new azure.sql.SqlServer("sqlserver", {
+    resourceGroupName: resourceGroup.name,
+
+    administratorLogin: "manualadmin",
+    administratorLoginPassword: "WeWontUseTh1sPwd",
```
lukehoban: Just to encourage best practices - should we construct this as a random password using `random.RandomString`?

azure-ts-msi-keyvault-rbac/index.ts
```diff
@@ -0,0 +1,170 @@
...
+// Azure SQL Database that we want to access from the application
+const database = new azure.sql.Database("sqldb", {
+    resourceGroupName: resourceGroup.name,
+    serverName: sqlServer.name,
+    requestedServiceObjectiveName: "S0"
+});
+
+// The connection string that has no credentials in it: authertication will come through MSI
+const connectionString = pulumi.all([sqlServer.name, database.name]).apply(([server, db]) => 
+    `Server=tcp:${server}.database.windows.net;Database=${db};`);
```
lukehoban: Note - we can simplify this now to:

```ts
const connectionString = pulumi.interpolate`Server=tcp:${sqlServer.name}.database.windows.net;Database=${database.name};`
```

azure-ts-msi-keyvault-rbac/index.ts
```diff
@@ -0,0 +1,170 @@
...
+const clientConfig = pulumi.output(azure.core.getClientConfig({}));
+const tenantId = clientConfig.apply(c => c.tenantId);
+
+// Determine the ID of the current user or service principal
+const currentPrincipal = clientConfig.apply(c => 
+    c.servicePrincipalObjectId !== "" 
+    // Current, only service principal ID is available in the context
+    ? c.servicePrincipalObjectId
+    // If logged in with a user, find their ID via Azure CLI
+    : <string>JSON.parse(execSync("az ad signed-in-user show --query objectId").toString()));
```
lukehoban: Interesting.  This is not going to be robust in 100% of cases, but I kind of like it to highlight how to pragmatically do "awkward" things like this.

azure-ts-msi-keyvault-rbac/index.ts
```diff
@@ -0,0 +1,170 @@
...
+
+// Key Vault to store secrets (e.g. Blob URL with SAS)
+const vault = new azure.keyvault.KeyVault("vault", {
+    resourceGroupName: resourceGroup.name,
+    sku: {
+        name: "standard",
+    },
+    tenantId: tenantId,
+    // It won't setup any access if pulumi is executed under user account
```
lukehoban: Is this still true even with the `execSync` part above?

azure-ts-msi-keyvault-rbac/index.ts
```diff
@@ -0,0 +1,170 @@
...
+    type: "block",
+
+    content: new pulumi.asset.FileArchive("./webapp/bin/Debug/netcoreapp2.2/publish")
+});
+
+const clientConfig = pulumi.output(azure.core.getClientConfig({}));
+const tenantId = clientConfig.apply(c => c.tenantId);
+
+// Determine the ID of the current user or service principal
+const currentPrincipal = clientConfig.apply(c =>
```
lukehoban: Does it make sense to create a new SP programmatically and use that instead?  Is there a need to use the actively running Principal?

azure-ts-msi-keyvault-rbac/index.ts
```diff
@@ -0,0 +1,170 @@
...
+        tenantId,
+        objectId: currentPrincipal,
+        secretPermissions: ["delete", "get", "list", "set"],
+    }]        
+});
+
+// Put the URL of the zip Blob to KV
+const secret = new azure.keyvault.Secret("deployment-zip", {
+    keyVaultId: vault.id,
+    value: azure.storage.signedBlobReadUrl(blob, storageAccount),
```
lukehoban: Do you know if this `value` is visibile in plain text in the checkpoint?  It might be nice to wrap `signedBlobReadUrl` in `pulumi.secret` (or whatever the API call is to turn an output into a secret), so that this gets encrypted in the checkpoint.  In general, it might be nice to double check that the sensitive values here are not replicated into the checkpoint if possible - it's not strictly critical for this example - but given this example is about trying to be more careful with access to secrets, also having it be the case that no secrets leak into the checkpoint would be a nice addition.

azure-ts-msi-keyvault-rbac/index.ts
mikhailshilkov: Totally, great hint!

azure-ts-msi-keyvault-rbac/index.ts
mikhailshilkov: Yes, this is based on https://github.com/terraform-providers/terraform-provider-azurerm/issues/3234 (I'm adding the link to this issue into comments)

azure-ts-msi-keyvault-rbac/index.ts
mikhailshilkov: The actively running principal has to be granted permissions to Key Vault so that it can actually add and then remove secrets from the Key Vault. Otherwise, `pulumi up` operations will fail.

azure-ts-msi-keyvault-rbac/index.ts
mikhailshilkov: Ouch, this must be a left-over; the workaround above fixes this issue. Removing.

azure-ts-msi-keyvault-rbac/index.ts
mikhailshilkov: Yes, it is visible in `pulumi stack export`. However, lots of things are there, e.g. for each storage account there is a `primaryAccessKey` sitting right there, or for an App Service 
```
"siteCredential": {
     "password": "redacted",
     "username": "$app902e9f6a"
},
```
The random string serving as password from the above is visible too.

Given that, the only reason to use `pulumi.secret` here seems to be to educate people it exists?

CyrusNajmabadi:
FYI, when you force push, it makes keeping up with the PR hard.  Just do normal pushes.  We squash anyways when finally comitting, so it ends up the same in github.  Thanks!

azure-ts-msi-keyvault-rbac/index.ts
CyrusNajmabadi: fwiw, this code just looks hard to read.  Consider just tweaking into:

```ts
const currentPrincipal = clientConfig.apply(c =>
    // Currently, only service principal ID is available in the context.  If we are provided the principle in the config, then
    // just use it.  Otherwise, if logged in with a user, find their ID via Azure CLI.
    // see https://github.com/terraform-providers/terraform-provider-azurerm/issues/3234
    c.servicePrincipalObjectId !== ""
        ? c.servicePrincipalObjectId
        : <string>JSON.parse(execSync("az ad signed-in-user show --query objectId").toString()));
```

azure-ts-msi-keyvault-rbac/webapp/Program.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
```
CyrusNajmabadi: do we have copyright in our example files?

azure-ts-msi-keyvault-rbac/webapp/Startup.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace webapp
+{
+
```
CyrusNajmabadi: remove blank line.

azure-ts-msi-keyvault-rbac/webapp/Program.cs
mikhailshilkov: No `*.cs` files do, so far. Some `*.ts` have them, some not.

azure-ts-msi-keyvault-rbac/index.ts
mikhailshilkov: Tweaked, thanks!

azure-ts-msi-keyvault-rbac/index.ts
lukehoban: Got it - makes sense - will be worth looking into those more in the future independent of this PR.

lukehoban:
@mikhailshilkov I think we can merge this now.   If there's any more feedback we can incorporate later.

mikhailshilkov:
Merging and starting a blog post on the subject

PR #306, opened 2019-05-29
Azure Static Website

mikhailshilkov:
Static websites on Azure Blob Storage, as discussed in https://github.com/pulumi/pulumi-azure/issues/260


azure-ts-static-website/README.md
```diff
@@ -0,0 +1,56 @@
+[![Deploy](https://get.pulumi.com/new/button.svg)](https://app.pulumi.com/new)
+
+# A Static Website Hosted on Azure Blob Storage + Azure CDN
+
+This example configures [Static website hosting in Azure Storage](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blob-static-website). One complication is the fact that the Static Website feature of Storage Accounts is not part of Azure Resource Manager, and is not configurable directly via Pulumi Azure provider.
+
+As a workaround, a custom dynamic provider and a dynamic resource are created. The provider delegates the setup to Azure CLI commands, while still providing Pulumi experience and lifecycle management.
```
lukehoban: Perhaps link "dynamic provider" to https://pulumi.io/reference/programming-model/#dynamicproviders?

azure-ts-static-website/staticWebsite.ts
```diff
@@ -0,0 +1,64 @@
...
+    accountName: pulumi.Input<string>;
+}
+
+const accountNameProp = "accountName";
+
+// There's currently no way to enable the Static Web Site feature of a storage account via ARM
+// Therefore, we created a custom provider which wraps corresponding Azure CLI commands
+class StorageStaticWebsiteProvider implements pulumi.dynamic.ResourceProvider {
+
+    check = async (olds: any, news: any) => {
```
lukehoban: This is a slightly unusual way to write this in TypeScript.  There are two alternatives that might be more natural:
1. Make these normal methods on the class
2. Use an object literal instead of a class (and then just use the single instance of the object literal below instead of `new ...`)

See (e.g.) https://pulumi.io/reference/programming-model/#example-github-labels-rest-api for an example of the latter.

azure-ts-static-website/staticWebsite.ts
```diff
@@ -0,0 +1,64 @@
...
+        if (olds[accountNameProp] !== news[accountNameProp]) {
+            replaces.push(accountNameProp);
+        }
+
+        return { replaces: replaces };
+    }
+
+    create = async (inputs: any) => {
+        const { execSync } = require("child_process");
+        execSync("az extension add --name storage-preview", { stdio: "ignore" });
```
lukehoban: Note that since this is `async` you could fairly easily make this exec call asynchronously.  But I doubt it has any practical performance impact - so fine to keep as-is if it's too hard to make this async.

azure-ts-static-website/staticWebsite.ts
```diff
@@ -0,0 +1,64 @@
...
+        const accountName = inputs[accountNameProp];
+        const executeToJson = (command: string) => JSON.parse(execSync(command, { stdio: ["pipe", "pipe", "ignore"] }).toString());
+        const update = executeToJson(`az storage blob service-properties update --account-name "${accountName}" --static-website --404-document 404.html --index-document index.html`);
+        if (!update.staticWebsite.enabled) {
+            throw new Error(`Static website update failed: ${update}`);
+        }
+
+        const endpoint = executeToJson(`az storage account show -n "${accountName}" --query "primaryEndpoints.web"`);
+        const hostName = endpoint.replace("https://", "").replace("/", "");
+        const webContainerName = "$web";
```
lukehoban: Is there any public reference that can be linked here to document what the two lines above are doing?  (And honestly the whole of this function :-)).

azure-ts-static-website/wwwroot/404.html
```diff
@@ -0,0 +1,5 @@
+<html>
+  <body>
+    <h1>That's a 404! Still, from the Blob Storage.</h1>
+  </body>
+</html>
```
lukehoban: Nit - line endings :-).

azure-ts-static-website/staticWebsite.ts
mikhailshilkov: Yeah, I tried, but it wasn't worth the hassle. Apart from promisify thing, there is a difference in options of `exec` vs `execSync` which didn't help.

mikhailshilkov:
@lukehoban Thanks for your comments, they should be all addressed now.

PR #307, opened 2019-05-30
Add an example of a dynamic resource provider that adds a custom domain to an Azure CDN endpoint.

praneetloke:


mikhailshilkov:
This seems to be related to #306, although with a different focus. Is that something we could/should combine?

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
...
+/** The custom domain host of the CDN endpoint. */
+export let cdnCustomDomainResource: CDNCustomDomainResource | undefined;
+
+// Create an Azure Resource Group
+const resourceGroup = new azure.core.ResourceGroup("resourceGroup", {
+    location: location,
+});
+
+// Create an Azure resource (Storage Account) so we can point the CDN endpoint to it.
+const storageAccount = new azure.storage.Account("storage-account", {
```
mikhailshilkov: Storage account names can't have dashes in them

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,231 @@
...
+      tenantID = process.env["ARM_TENANT_ID"];
+      subscriptionID = process.env["ARM_SUBSCRIPTION_ID"];
+    }
+
+    // If they are still empty, then throw an error.
+    if (!clientID || !clientSecret || !tenantID || !subscriptionID) {
+      throw new Error("ARM credentials could not be detected.");
+    }
+
+    const credentials = await msRestAzure.loginWithServicePrincipalSecret(clientID, clientSecret, tenantID);
```
mikhailshilkov: This fails when I run it from CLI:

> Checking env vars for ARM credentials.
> Error: ARM credentials could not be detected.: Error: ARM credentials could not be detected.

Shouldn't we do something like `const cred = await msRestAzure.AzureCliCredentials.create();` instead?

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
...
+        "application/javascript",
+        "application/json",
+        "application/xml",
+        "image/png",
+        "image/jpeg"
+    ],
+    origins: [
+        {
+            name: "cdn-origin",
+            hostName: storageAccount.primaryBlobEndpoint,
```
mikhailshilkov: This needs to be a hostname like `bla.blob.core.windows.net` not the full URL like `https://bla.blob.core.windows.net`. Consider using something like `storageAccount.primaryBlobEndpoint.apply(ep => url.parse(ep).hostname!)`. Applies to `originHostHeader` too.

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
...
+export const cdnEndpointUrl = pulumi.interpolate `https://${cdnEndpoint.hostName}`;
+
+pulumi.all([resourceGroup.name, cdnProfile.name, cdnEndpoint.name])
+    .apply(([resourceGroupName, cdnProfileName, cdnEndpointName]) => {
+        cdnCustomDomainResource = new CDNCustomDomainResource("cdnCustomDomain", {
+            // Ensure that there is a CNAME record for mycompany.com
+            // pointing to my-cdn-endpoint.azureedge.net.
+            // You would do that in your domain registrar's portal.
+            customDomainHostName: "mycompany.com",
+            customDomainName: "custom-domain",
```
mikhailshilkov: Why do we need this property? Could we just use `name` internally instead?

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
...
+            profileName: cdnProfileName,
+            endpointName: cdnEndpointName,
+            // This will enable HTTPS through Azure's one-click
+            // automated certificate deployment.
+            // The certificate is fully managed by Azure from provisioning
+            // to automatic renewal at no additional cost to you.
+            httpsEnabled: true,
+            resourceGroupName: resourceGroupName
```
mikhailshilkov: Nit: missing commas here and in other places

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+// Create an Azure Resource Group
+const resourceGroup = new azure.core.ResourceGroup("resourceGroup", {
+    location: location,
+});
+
+// Create an Azure resource (Storage Account) so we can point the CDN endpoint to it.
+const storageAccount = new azure.storage.Account("storage-account", {
+    resourceGroupName: resourceGroup.name,
+    location: resourceGroup.location,
```
mikhailshilkov: You shouldn't pass the location explicitly: it will be auto-filled from resource group properties.

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
...
+            name: "cdn-origin",
+            hostName: storageAccount.primaryBlobEndpoint,
+            httpsPort: 443,
+        }
+    ]
+});
+
+export const cdnEndpointUrl = pulumi.interpolate `https://${cdnEndpoint.hostName}`;
+
+pulumi.all([resourceGroup.name, cdnProfile.name, cdnEndpoint.name])
```
mikhailshilkov: I think it would look more natural if we accepted `Input<T>` in `CustomDomainOptions` and then on the client side it would be just a constructor call without `pulumi.all`.

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
...
+// Create an Azure resource (Storage Account) so we can point the CDN endpoint to it.
+const storageAccount = new azure.storage.Account("storage-account", {
+    resourceGroupName: resourceGroup.name,
+    location: resourceGroup.location,
+    accountTier: "Standard",
+    accountReplicationType: "LRS",
+});
+
+// Create a Blob container in the storage account.
+const blobContainer = new azure.storage.Container("blob-container", {
```
mikhailshilkov: Consider documenting (in README and/or inline comments) where the static website files must be put and which URL to use to access them through CDN.

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as azure from "@pulumi/azure";
+
+import { CDNCustomDomainResource } from "./cdnCustomDomain";
+
+const location = "West US";
+
+/** The custom domain host of the CDN endpoint. */
+export let cdnCustomDomainResource: CDNCustomDomainResource | undefined;
```
mikhailshilkov: What is the reason to export it?

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as azure from "@pulumi/azure";
+
+import { CDNCustomDomainResource } from "./cdnCustomDomain";
+
+const location = "West US";
```
mikhailshilkov: Consider using a configuration parameter for the location instead of a hard-coded value.

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,86 @@
...
+const blobContainer = new azure.storage.Container("blob-container", {
+    resourceGroupName: resourceGroup.name,
+    storageAccountName: storageAccount.name,
+    containerAccessType: "blob"
+});
+
+const cdnProfile = new azure.cdn.Profile("cdn-profile", {
+    location: location,
+    resourceGroupName: resourceGroup.name,
+    sku: "Standard_Akamai",
```
mikhailshilkov: What are the implications of choosing Akamai instead of Microsoft?

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,231 @@
...
+    const cdnClient = await this.getCDNManagementClient();
+    const result =
+    await cdnClient.customDomains
+      .create(
+        inputs.resourceGroupName,
+        inputs.profileName,
+        inputs.endpointName,
+        inputs.customDomainName,
+        inputs.customDomainHostName);
+    console.log(`Custom domain provisioning state is ${ result.resourceState }.`);
```
mikhailshilkov: Do we want to keep `console.log` in the examle?

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: The dynamic provider tries to use the Pulumi CLI's Azure config, and if not looks for environment vars. I haven't tried using `msRestAzure.AzureCliCredentials.create();` myself, to be honest. I did it this way since I typically use a Service Connection in Azure DevOps and don't use the Az CLI in any of my tasks.

azure-ts-dynamicresource/index.ts
praneetloke: Ah yes. This is the result of me trying to change the names I used in my personal project to something generic. I'll change this.

azure-ts-dynamicresource/index.ts
praneetloke: `primaryBlobEndpoint` actually does work for both of those properties. But if you think mapping the hostname would be better, I can change it.

mikhailshilkov: Hmm, I'm getting errors running that... weird

azure-ts-dynamicresource/cdnCustomDomain.ts
mikhailshilkov: The README suggests `az login`, so it should work. Also, that's probably what users will use while trying examples.

azure-ts-dynamicresource/index.ts
praneetloke: Yep. Absolutely. I can change the provider to use the name of the custom resource as the "name" of the custom domain as well. Also, even though it is sort of a display name the Azure portal doesn't seem to show it anywhere.

azure-ts-dynamicresource/index.ts
praneetloke: Ah gotcha. Removed.

azure-ts-dynamicresource/index.ts
praneetloke: Yes, I definitely agree. In fact, that is what I started with. But if you look closely at the formal parameters of the functions in my dynamic resource provider, I used the type `CustomDomainOptions` for `inputs` instead of `any`. That means I will need to deal with an `Input<T>` inside those functions, which I couldn't see a way to unwrap them there without returning a `pulumi.Output<T>`, which I couldn't find a cleaner way to transform. Things got really confusing there to be honest.

I see that in your static website example, you are retaining the type of the `inputs` formal param as `any` and are accepting an `Input<T>` for your provider input. I am not sure how when your `create()` is called, that you are getting a regular string instead of an `Input<T>`. In any case, I would prefer to have the `inputs` param typed concretely instead of having to treat it as an anonymous object.

mikhailshilkov: Great point! I stitched my example from several other implementations, and the reason behind `any` is exactly as you say. Maybe @jen20 can clarify how `Input`s get resolved and where we stand on using types everywhere?

azure-ts-dynamicresource/index.ts
praneetloke: To show that the dynamic resource's outputs are exportable too.

azure-ts-dynamicresource/index.ts
praneetloke: I'll leave it as-is, but will add a comment that the user should consider using the Pulumi config system to externalize it.

azure-ts-dynamicresource/index.ts
praneetloke: Will do.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: I'd like to since the custom domain and HTTPS provisioning can, in some cases, not be immediately active.

azure-ts-dynamicresource/index.ts
praneetloke: I'll add a link to https://docs.microsoft.com/en-us/azure/cdn/cdn-features.

azure-ts-dynamicresource/index.ts
praneetloke: Yeah I wonder if I could do a union type here for my args like: `pulumi.Input<T> | string`, and then still be able to pass an `Input<T>` from the outside, but within my dynamic resource use them as regular strings?

praneetloke:
> This seems to be related to #306, although with a different focus. Is that something we could/should combine?

I somehow missed this comment of yours and see that your PR is now merged. In any case, FWIW, I would have replied that I would not like to merge the two examples. I'd like to keep them as distinct ones so it doesn't look like one large example. Though, I do agree that the two go together very well.

azure-ts-dynamicresource/index.ts
praneetloke: Ah sorry. I compared this with the wrong project of mine. In my personal project, I am using the function app's host name as the origin. When I translated this to use the storage account, I used the wrong property. I should have used `primaryBlobHost`.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Good catch! I have updated this to try and get the credentials from the CLI if the env vars are not set.

mikhailshilkov:
I'm getting an error during `cdnCustomDomain` preview/creation:

```
Error: Error "SyntaxError: Unexpected token o in JSON at position 1" occurred while parsing the response body - [object Object].: Error: Error "SyntaxError: Unexpected token o in JSON at position 1" occurred while parsing the response body - [object Object].
        at new RestError (C:\work\pulumi\examples\azure-ts-dynamicresource\node_modules\@azure\ms-rest-js\lib\restError.ts:18:5)
        at errorHandler (C:\work\pulumi\examples\azure-ts-dynamicresource\node_modules\@azure\ms-rest-js\lib\policies\deserializationPolicy.ts:196:15)
        at process._tickCallback (internal/process/next_tick.js:68:7)

  pulumi-nodejs:dynamic:Resource (azure:cdn:Endpoint:CustomDomains:cdnCustomDomain):
    error: Error "SyntaxError: Unexpected token o in JSON at position 1" occurred while parsing the response body - [object Object].
```

Running from CLI, other resources are already created. Any ideas?

azure-ts-dynamicresource/index.ts
mikhailshilkov: So what we can do is to create two distinct types: an internal `Props` (name not final) which has no inputs in it and an exported `Args` which has the same shape but has all inputs. `Props` will be used in the dynamic provider implementation and `Args` is like normal args in all our components - to be passed into the constructor.

In order to avoid copy-pasting and typos, we can define a mapped type. I tried it on my static website example:

``` ts
// A mapped type to convert any interface into a type of the same shape but all types converted from T to Input<T>
type Inputs<T> = {
    [P in keyof T]: pulumi.Input<T[P]>;
}
  
interface StorageStaticWebsiteProps {
    accountName: string;
}

export type StorageStaticWebsiteArgs = Inputs<StorageStaticWebsiteProps>;
```

I was able to replace all `any` with `StorageStaticWebsiteProps`. Works like a charm.

@CyrusNajmabadi @lukehoban Don't we have a type like `Inputs` above in the standard Pulumi library? Do we what one?

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,244 @@
...
+
+export interface CustomDomainOptions {
+  resourceGroupName: string;
+  profileName: string;
+  endpointName: string;
+  customDomainHostName: string;
+  httpsEnabled: boolean;
+}
+
+export interface CustomDomainOutputs extends cdnManagement.CdnManagementModels.CustomDomainsCreateResponse {
```
mikhailshilkov: Is this an internal structure to bring the state between multiple runs or will a user interact with it? If the former, why is it exported? If the latter, its shape seems to be different from what a typical component would expose as outputs? Is there guidance for dynamic providers in this regards?

praneetloke:
> Running from CLI, other resources are already created. Any ideas?

Yes, see https://github.com/Azure/azure-sdk-for-js/issues/2842#issuecomment-498895303. MS just fixed an issue that crept into their SDK when they upgraded `axios` to `0.19.0` to patch a vulnerability.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: > Is this an internal structure to bring the state between multiple runs or will a user interact with it? If the former, why is it exported?

I exported it because I am under the assumption that the type of the outputs objects could be used by the caller to extract specific properties. But I am not sure if it is necessary, and if there is a more "accepted" way for dynamic providers to declare an output type or if one is even needed. At the very least I could un-export this type.

> Is there guidance for dynamic providers in this regards?

I am going to speak to someone here at Pulumi about dynamic provider outputs too and have an answer for this.

azure-ts-dynamicresource/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "azure-ts-appservice",
+    "version": "1.0.0",
+    "devDependencies": {
+        "@types/node": "^10.3.3"
+    },
+    "dependencies": {
+        "@azure/arm-cdn": "^4.2.0",
+        "@azure/ms-rest-nodeauth": "^1.1.1",
```
mikhailshilkov: Shall we set the versions to the ones that don't cause the error?

azure-ts-dynamicresource/README.md
```diff
@@ -0,0 +1,85 @@
...
+            customDomainName: "custom-domain",
+            profileName: cdnProfileName,
+            endpointName: cdnEndpointName,
+            // This will enable HTTPS through Azure's one-click
+            // automated certificate deployment.
+            // The certificate is fully managed by Azure from provisioning
+            // to automatic renewal at no additional cost to you.
+            httpsEnabled: true,
+            resourceGroupName: resourceGroupName
+        }, { parent: cdnEndpoint });
```
CyrusNajmabadi: nit: we would dedent this all twice.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,244 @@
+import * as pulumi from "@pulumi/pulumi";
```
CyrusNajmabadi: copyright?

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,244 @@
...
+import * as azure from "@pulumi/azure";
+
+import * as msRestAzure from "@azure/ms-rest-nodeauth";
+import * as cdnManagement from "@azure/arm-cdn";
+import { ServiceClientCredentials } from "@azure/ms-rest-js";
+
+export interface CustomDomainOptions {
+  resourceGroupName: string;
```
CyrusNajmabadi: nit: you have 2-space indent on.  But i think all our sdk/example code is 4-space.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,244 @@
...
+      };
+    } else {
+      result = {
+        inputs: news
+      };
+    }
+
+    return result;
+  }
+  async diff(id: string, previousOutput: CustomDomainOutputs, news: CustomDomainOptions): Promise<pulumi.dynamic.DiffResult> {
```
CyrusNajmabadi: newline above this.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,244 @@
...
+        newInputs.profileName,
+        newInputs.endpointName,
+        this.name);
+
+    currentOutputs.inputs.httpsEnabled = true;
+    return {
+      outs: currentOutputs
+    };
+  }
+}
```
CyrusNajmabadi: note: i actually skipped most of this as i don't feel qualified to review a provider.  perhaps w can get @pgavlin to take a look.  It's something i want to ramp up on, so i'll read this and ask questions.  But i'd like someone more informed to take a deeper look.

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,96 @@
+import * as pulumi from "@pulumi/pulumi";
```
CyrusNajmabadi: copyright.

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const cdnEndpoint = new azure.cdn.Endpoint("my-cdn-endpoint", {
+    name: "my-cdn-endpoint",
+    resourceGroupName: resourceGroup.name,
+    profileName: cdnProfile.name,
+    isHttpsAllowed: true,
+    isHttpAllowed: false,
+    isCompressionEnabled: true,
+    originHostHeader: storageAccount.primaryBlobHost,
+    contentTypesToCompresses: [
```
CyrusNajmabadi: aside: contentTypesToCompresses is a bizarro name.  "Compresses"??

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,96 @@
...
+            // You would do that in your domain registrar's portal.
+            customDomainHostName: "mycompany.com",
+            profileName: cdnProfileName,
+            endpointName: cdnEndpointName,
+            // This will enable HTTPS through Azure's one-click
+            // automated certificate deployment.
+            // The certificate is fully managed by Azure from provisioning
+            // to automatic renewal at no additional cost to you.
+            httpsEnabled: true,
+        }, { parent: cdnEndpoint });
```
CyrusNajmabadi: i'm a little confused why any of this would be needed.  can't you just write:

```ts
cdnCustomDomainResource = new ...(... {
    resourceGroupName: resourceGroup.name,
    // etc.?
}
```

I don't see why you need the .all or .apply.

azure-ts-dynamicresource/index.ts
praneetloke: Yes. Unfortunately, that's how the terraform property is named. It is a typo in the provider itself.

azure-ts-dynamicresource/index.ts
praneetloke: The reason is it isn't clear from the docs about the dynamic provider if one can pass `Input<T>` props to the dynamic resource and still get strong typing in the provider's functions for the inputs.

praneetloke:
I just spoke to @pgavlin about the dynamic provider and I am going to make a bunch of changes based on that, as well as update our dynamic providers to reflect the recommendations.

CyrusNajmabadi:
Great.  Feel free to ping again when ready!

praneetloke:
@mikhailshilkov @CyrusNajmabadi @pgavlin can you please take another look now? I have cleaned-up a lot of things. The main difference is, I am now accepting `Input<T>` for the appropriate properties. Declared a separate interface that is local to the dynamic provider itself so I still get strong typing when dealing with the inputs in the CRUD functions. This is what was recommended by Pat.

Also, as recommended, I have declared the outputs as `public readonly` members with an `Output<T>` type in the `dynamic.Resource` class itself.

I will make an update to our dynamic providers docs based on this. I have added comments for most things in this PR.

azure-ts-dynamicresource/README.md
CyrusNajmabadi: not sure why this was resolved.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,276 @@
...
+        const validationOutput = await cdnClient.endpoints.validateCustomDomain(
+            news.resourceGroupName, news.profileName, news.endpointName, news.customDomainHostName);
+        if (!validationOutput.customDomainValidated) {
+            result = {
+                failures: [
+                    {
+                        property: this.name,
+                        reason: validationOutput.reason || "domain_validation_failed"
+                    }
+                ]
```
CyrusNajmabadi: consider trimming this down to:

```ts
            result = {
                failures: [{
                    property: this.name,
                    reason: validationOutput.reason || "domain_validation_failed",
                }],
```

(also in a few other cases).

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,276 @@
...
+            };
+        } else {
+            result = {
+                inputs: news
+            };
```
CyrusNajmabadi: single prop objects are commonly written as `result = { inputs: news };`

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,276 @@
...
+        if (previousOutput.httpsEnabled !== news.httpsEnabled) {
+            changes = true;
+            replaces.push("httpsEnabled");
+        }
+
+        return {
+            deleteBeforeReplace: deleteBeforeReplace,
+            replaces: replaces,
+            changes: changes
+        };
```
CyrusNajmabadi: can be simplified to just: `return { deleteBeforeReplace, replaces, changes };`

azure-ts-dynamicresource/cdnCustomDomain.ts
CyrusNajmabadi: assigning to result just to return below seems unnecessary.  you can just `return { inputs: news };`

azure-ts-dynamicresource/README.md
praneetloke: Oh oops. Didn't mean to.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,276 @@
...
+    async create(inputs: DynamicProviderInputs): Promise<pulumi.dynamic.CreateResult> {
+        const cdnClient = await this.getCDNManagementClient();
+        const result =
+            await cdnClient.customDomains
+                .create(
+                    inputs.resourceGroupName,
+                    inputs.profileName,
+                    inputs.endpointName,
+                    this.name,
+                    inputs.customDomainHostName);
```
CyrusNajmabadi: heavy wrapping.  consider just:

```ts
const result = await cdnClient.customDomains.create(
    inputs.resourceGroupName,
    inputs.profileName,
    inputs.endpointName,
    this.name,
    inputs.customDomainHostName);
```

Or, if it's not too long:

```ts
const result = await cdnClient.customDomains.create(
    inputs.resourceGroupName, inputs.profileName, inputs.endpointName, this.name, inputs.customDomainHostName);
```

CyrusNajmabadi: Also, in general, in pulumi resources and whatnot, we have 'name' come first.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,276 @@
...
+
+        if (inputs.httpsEnabled) {
+            console.log("Enabling HTTPS for the custom domain");
+
+            await cdnClient.customDomains
+                .enableCustomHttps(
+                    inputs.resourceGroupName,
+                    inputs.profileName,
+                    inputs.endpointName,
+                    this.name);
```
CyrusNajmabadi: same comments for this.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,276 @@
...
+            name: this.name,
+            resourceGroupName: inputs.resourceGroupName,
+            profileName: inputs.profileName,
+            endpointName: inputs.endpointName,
+            customDomainHostName: inputs.customDomainHostName,
+            httpsEnabled: inputs.httpsEnabled,
+        };
+
+        return {
+            id: result.id!,
```
CyrusNajmabadi: not clear why the `!` is necessary.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,276 @@
...
+        const cdnClient = await this.getCDNManagementClient();
+        const result = await cdnClient.customDomains.deleteMethod(props.resourceGroupName, props.profileName, props.endpointName, this.name);
+        if (result._response.status >= 400) {
+            throw new Error("Error response received while trying to delete the custom domain.");
+        }
+
+        if (!result.resourceState) {
+            return;
+        }
+        if (result.resourceState !== 'Deleting') {
```
CyrusNajmabadi: can you have a blank line above this for consistency.  or remove hte blank line between the first two ifs.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,276 @@
...
+     */
+    async update(id: string, currentOutputs: DynamicProviderOutputs, newInputs: DynamicProviderInputs): Promise<pulumi.dynamic.UpdateResult> {
+        const cdnClient = await this.getCDNManagementClient();
+        if (newInputs.httpsEnabled) {
+            await cdnClient.customDomains
+                .enableCustomHttps(
+                    newInputs.resourceGroupName,
+                    newInputs.profileName,
+                    newInputs.endpointName,
+                    this.name);
```
CyrusNajmabadi: same comments about wrapping.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,276 @@
...
+                outs: currentOutputs
+            };
+        }
+
+        await cdnClient.customDomains
+            .disableCustomHttps(
+                newInputs.resourceGroupName,
+                newInputs.profileName,
+                newInputs.endpointName,
+                this.name);
```
CyrusNajmabadi: same comment.

azure-ts-dynamicresource/index.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * The location where our resource group and the resources under it will be created.
+ *
+ * To externalize this value, and make this configurable across environments/stacks,
+ * learn more at https://pulumi.io/reference/config/.
+ */
+const location = "West US";
+
+/** The custom domain host of the CDN endpoint. */
+export let cdnCustomDomainResource: CDNCustomDomainResource | undefined;
```
CyrusNajmabadi: why declare this up here.  you can just move down to the place you actually create it.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Done (and in a few other places)

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Sure. I can change it.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Really? Those are keys in the return type. Could I still do what you are suggesting?

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Sure I did this here and in a couple of other places.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: It is an optional in the `result`. I get a lint error for not checking for null-ness otherwise. Actually, I have a question for @pgavlin here. Should I be setting the value of the `id` in the output returned by the `create()`?

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Done.

azure-ts-dynamicresource/index.ts
praneetloke: Yep. Moved it.

azure-ts-dynamicresource/README.md
praneetloke: It's updated now.

azure-ts-dynamicresource/package.json
praneetloke: Done.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,258 @@
...
+    customDomainHostName: pulumi.Input<string>;
+    httpsEnabled: boolean;
+}
+
+/**
+ * DynamicProviderInputs represents the inputs that are passed as inputs
+ * to each function in the implementation of a `pulumi.dynamic.ResourceProvider`.
+ * The property names in this must match `CustomDomainOptions`.
+ */
+interface DynamicProviderInputs {
```
mikhailshilkov: Don't you like my suggested trick with a mapped type?

praneetloke: Ah it got buried under my updates. I'll reply to that comment.

azure-ts-dynamicresource/index.ts
praneetloke: This is an elegant solution and one that I prefer, but would actually love it if it was part of the dynamic providers library itself. Aside from that, I do feel mapped types are sort of an advanced concept in TS for most beginners who are using TS, let alone Pulumi. That's just my opinion, though. I'd love to see what @lukehoban thinks about this.

azure-ts-dynamicresource/cdnCustomDomain.ts
CyrusNajmabadi: yup.  in js/ts  `{ a, b, c }` is shorthand for `{ a: a, b: b, c: c }`.

azure-ts-dynamicresource/cdnCustomDomain.ts
lukehoban: > Should I be setting the value of the id in the output returned by the create()?

Yes - you'll want to ensure you return an `id` that represents the resource in the backing cloud provider.  This may be necessary for example to `delete` the resource later - as you'll need to use this `id` to look it up.

It's slightly concerning that the `id` is an optional return value from the cloud provider!

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Ohh...learned something new!

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: > It's slightly concerning that the id is an optional return value from the cloud provider!

Yes, agreed. Though, it seems that the custom domain resource does always get an ID. I believe this is an optional because the underlying CDN management client extends itself from another type `Resource`, which is where this `id` comes from and it would seem that it is optional for other resources. So I think we are safe in this instance.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: However, I'll leave it as-is in this case.

azure-ts-dynamicresource/README.md
```diff
@@ -0,0 +1,85 @@
+[![Deploy](https://get.pulumi.com/new/button.svg)](https://app.pulumi.com/new)
+
+# Azure CDN Custom Domain Dynamic Provider
+
+## Prerequisites
+
+Prior to running this example, ensure that the custom domain, you will be adding to your CDN endpoint, has a CNAME record in your domain registrar's DNS management portal. The dynamic provider will validate that the mapping exists and will fail the deployment if the CDN endpoint is not reachable through your custom domain.
```
pgavlin: nit: extra commas

> Prior to running this example, ensure that the custom domain you will be adding to your CDN endpoint has a CNAME record in your domain registrar's DNS management portal

azure-ts-dynamicresource/README.md
```diff
@@ -0,0 +1,85 @@
...
+    Update duration: ...
+    ```
+
+## Dynamic Providers
+
+Learn more about dynamic providers [here](https://pulumi.io/reference/programming-model/#dynamicproviders).
+
+## Known Issues
+
+If you get a 404 error when deleting a custom domain or a deserialization error `SyntaxError: Unexpected token o in JSON at position 1`, this was due to a bug in the Azure nodeJS SDK. See https://github.com/Azure/azure-sdk-for-js/issues/2842. Ensure you have the latest versions by running `npm list @azure/ms-rest-js @azure/ms-rest-azure-js`. The versions you should have are:
```
pgavlin: reword

> Bugs in older versions of the Azure NodeJS SDK can cause 404 errors or deserialization errors during some operations (see https://github.com/Azure/azure-sdk-for-js/issues/2842 for more details). You can ensure that you have the latest versions of the relevant SDK packages by running...

azure-ts-dynamicresource/README.md
```diff
@@ -0,0 +1,85 @@
...
+- `@azure/ms-rest-js` - 1.8.10
+- `@azure/ms-rest-azure-js` - 1.3.7
+
+If you do not have the latest versions, then delete the `node_modules` folder, and the `package-lock.json`, or if you are using yarn, then `yarn-lock.json`, and then run `npm install` (or `yarn install`) again. Then verify that you have the correct versions again by running the `npm list ...` command.
```
pgavlin: > If you do not have the latest versions, delete your `node_modules` folder and `package-lock.json` (or `yarn.lock` if you are using yarn) and run `npm install` (or `yarn install`) again.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,258 @@
...
+ * CustomDomainOptions represents the inputs to the dynamic resource.
+ * Any property of type `Input<T>` will automatically be resolved by the custom dynamic resource
+ * before passing them to the dynamic resource provider.
+ */
+export interface CustomDomainOptions {
+    resourceGroupName: pulumi.Input<string>;
+    profileName: pulumi.Input<string>;
+    endpointName: pulumi.Input<string>;
+    customDomainHostName: pulumi.Input<string>;
+    httpsEnabled: boolean;
```
pgavlin: `pulumi.Input<boolean>`?

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,258 @@
...
+        return { inputs: news };
+    }
+
+    async diff(id: string, previousOutput: DynamicProviderOutputs, news: DynamicProviderInputs): Promise<pulumi.dynamic.DiffResult> {
+        const replaces: string[] = [];
+        let changes = false;
+        let deleteBeforeReplace = false;
+
+        if (previousOutput.customDomainHostName !== news.customDomainHostName || previousOutput.name !== this.name) {
+            console.warn("Changing the domain name properties will cause a downtime.")
```
pgavlin: `s/cause a downtime/cause downtime`

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        if (previousOutput.endpointName !== news.endpointName) {
+            deleteBeforeReplace = true;
+            replaces.push("endpointName");
+        }
+
+        // HTTPS can be enabled/disabled in-place.
+        if (previousOutput.httpsEnabled !== news.httpsEnabled) {
+            changes = true;
```
pgavlin: Changes really ought to be true in all cases that there are changes, even if those changes are already listed in `replaces`.

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,258 @@
...
+            throw new Error("Error response received while trying to delete the custom domain.");
+        }
+        if (!result.resourceState) {
+            return;
+        }
+        if (result.resourceState !== 'Deleting') {
+            throw new Error(`Provisioning state of the custom domain was expected to be 'Deleting', but was ${result.resourceState}.`);
+        }
+
+        console.log("The request to delete was successful. However, it can take a minute or two to fully complete deletion.");
```
pgavlin: FWIW, the console logging is pretty unusual for a provider. I think it makes more sense for these to be ephemeral log messages (like the k8s status messages), but I can't recall offhand how to do that.

azure-ts-dynamicresource/cdnCustomDomain.ts
mikhailshilkov: Why? I agree it reads redundant.

azure-ts-dynamicresource/index.ts
mikhailshilkov: from @ericrudder in another example:
>it woudl be great if we could "unhard code these" everywhere ... or at least not grow debt
> it's really a pulumi aniti-pattern, and we should avoid these ... we have a few others (like poor password usage) and they bite us when users copy/paste

azure-ts-dynamicresource/cdnCustomDomain.ts
```diff
@@ -0,0 +1,258 @@
...
+        const cdnClient = await this.getCDNManagementClient();
+
+        // If none of the CDN properties changed, then there is nothing to be validated.
+        if (olds.profileName === news.profileName &&
+            olds.endpointName === news.endpointName &&
+            olds.customDomainHostName === news.customDomainHostName) {
+            return { inputs: news };
+        }
+
+        const validationOutput = await cdnClient.endpoints.validateCustomDomain(
```
mikhailshilkov: When I run preview on a clean stack, I get an error (probably, because the resource group doesn't exist yet):

```
    Error: Resource group '04da6b54-80e4-46f7-96ec-b56ff0331ba9' could not be found.: Error: Resource group '04da6b54-80e4-46f7-96ec-b56ff0331ba9' could not be found.
        at new RestError (C:\work\pulumi\examples\azure-ts-dynamicresource\node_modules\@azure\ms-rest-js\lib\restError.ts:18:5)
        at C:\work\pulumi\examples\azure-ts-dynamicresource\node_modules\@azure\ms-rest-js\lib\policies\deserializationPolicy.ts:117:27
        at process._tickCallback (internal/process/next_tick.js:68:7)
```

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: I'll leave it up to the reader to make any such changes since it is more of a code style optimization. I don't think the example needs to be as succinct as possible.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: I intentionally left it since its value is derived from a pulumi config property. Would you still recommend that I have it as an `Input<boolean>`? How about a union type `Input<boolean> | boolean`?

mikhailshilkov: `Input` is itself a union: `Input<T> = T | Promise<T> | OutputInstance<T>` so `Input<boolean>` includes `boolean`.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Ah gotcha I'll change it. Reading the doc for [`diff`](https://pulumi.io/reference/programming-model/#diff-id-olds-news) again, I now see what it means, that it is regardless of the type of update; replace vs delete/create.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: `pulumi.log.info()|warn()|...`? I see that those functions accept an ephemeral property. It'll be good to demonstrate that to the reader, that Pulumi has that capability. I'll switch these to that.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Hmm..does it pass when you re-run it?

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Ah yes I see it now. I've changed it.

azure-ts-dynamicresource/cdnCustomDomain.ts
praneetloke: Something is not quite right here. I am looking into this.

praneetloke: I have updated the provider to address this. The cause of this error was because `check` seems to be called during a preview as well, when the resource group doesn't exist yet. That is unfortunate, because the call to the validation of the custom domain seemed appropriate to have in the `check` function. I have moved it to the `create` function now. The `check` function is now just validating that there is a value for the required inputs.

When I used this for my own project originally, it was an existing app to which I added the custom domain, so the resource group et al. existed already, which is why I didn't notice this.

praneetloke:
I am planning to merge tomorrow, if there are no more comments on the core dynamic provider itself.

azure-ts-dynamicresource/README.md
praneetloke: Reworded it differently.

azure-ts-dynamicresource/README.md
praneetloke: Done.

azure-ts-dynamicresource/README.md
```diff
@@ -0,0 +1,85 @@
...
+    Update duration: ...
+    ```
+
+## Dynamic Providers
+
+Learn more about dynamic providers [here](https://pulumi.io/reference/programming-model/#dynamicproviders).
+
+## Known Issues
+
+If you get a 404 error when deleting a custom domain or a deserialization error `SyntaxError: Unexpected token o in JSON at position 1`, these were due to bugs in the older versions of the Azure nodeJS SDK. See https://github.com/Azure/azure-sdk-for-js/issues/2842 for more info. Ensure that you have the latest versions by running `npm list @azure/ms-rest-js @azure/ms-rest-azure-js`. The versions you should have are:
```
pgavlin: nit: `NodeJS`

PR #308, opened 2019-05-30
AWS AppSync / GraphQL

mikhailshilkov:
Basic example of defining an AWS AppSync GraphQL endpoint from Pulumi


stack72:
This LGTM! Thanks for this:

```
▶ curl -XPOST -H "Content-Type:application/graphql" -H "x-api-key:$(pulumi stack output key)" -d '{ "query": "mutation AddTenant { addTenant(id: \"123\", name: \"FirstCorp\") { id name } }" }' "$(pulumi stack output endpoint)"
{"data":{"addTenant":{"id":"123","name":"FirstCorp"}}}%
```

PR #309, opened 2019-06-01
Add a couple Azure web server examples

joeduffy:
This adds two new examples:

1) A simple Azure web server in a VM, in TypeScript, with a public IP.

2) An Azure web server component that hides many details about how
   to create a VM with a public IP address, etc, and then uses it
   from the primary program to create a variable number of instances.

I used these in recent customer demos and found them to be helpful for
telling a story around components for those working with Azure.

### DIFF
--- a/azure-ts-webserver-component/index.ts
+++ b/azure-ts-webserver-component/index.ts
@@ -1,0 +1,21 @@
+import * as pulumi from "@pulumi/pulumi";
+import { WebServer } from "./webserver";
+
+// Get the desired username and password for our webserver VMs.
+let config = new pulumi.Config();
+let count = config.getNumber("count") || 2;
+let username = config.require("username");
+let password = config.requireSecret("password");
+
+// Now, allocate a few websever VMs -- by default, just 2, but this is configurable.
+export let ipAddresses = [];
+for (let i = 0; i < count; i++) {
+    let server = new WebServer(`ws-${i}`, {
+        username,
+        password,
+        bootScript: `#!/bin/bash\n
+echo "Hello, from Server #{i+1}!" > index.html
+nohup python -m SimpleHTTPServer 80 &`,
+    });
+    ipAddresses.push(server.getIpAddress());
+}
--- a/azure-ts-webserver-component/index.ts
+++ b/azure-ts-webserver-component/index.ts
@@ -3,33 +3,33 @@ import * as azure from "@pulumi/azure";
 import { WebServer } from "./webserver";
 
 // Get the desired username and password for our webserver VMs.
-let config = new pulumi.Config();
-let count = config.getNumber("count") || 2;
-let username = config.require("username");
-let password = config.requireSecret("password");
+const config = new pulumi.Config();
+const count = config.getNumber("count") || 2;
+const username = config.require("username");
+const password = config.requireSecret("password");
 
 // All resources will share a resource group.
-let resourceGroupName = new azure.core.ResourceGroup("server").name;
+const resourceGroupName = new azure.core.ResourceGroup("server-rg").name;
 
 // Create a network and subnet for all VMs.
-let network = new azure.network.VirtualNetwork("server-network", {
+const network = new azure.network.VirtualNetwork("server-network", {
     resourceGroupName,
-    addressSpaces: [ "10.0.0.0/16" ],
+    addressSpaces: ["10.0.0.0/16"],
     subnets: [{
         name: "default",
         addressPrefix: "10.0.1.0/24",
     }],
 });
-let subnet = new azure.network.Subnet("server-subnet", {
+const subnet = new azure.network.Subnet("server-subnet", {
     resourceGroupName,
     virtualNetworkName: network.name,
     addressPrefix: "10.0.2.0/24",
 });
 
 // Now, allocate a few websever VMs -- by default, just 2, but this is configurable.
-export let ipAddresses = [];
+export const ipAddresses = [];
 for (let i = 0; i < count; i++) {
-    let server = new WebServer(`ws-${i}`, {
+    const server = new WebServer(`ws-${i}`, {
         username,
         password,
         bootScript: `#!/bin/bash\n
--- a/azure-ts-webserver-component/index.ts
+++ b/azure-ts-webserver-component/index.ts
@@ -1,4 +1,5 @@
 import * as pulumi from "@pulumi/pulumi";
+import * as azure from "@pulumi/azure";
 import { WebServer } from "./webserver";
 
 // Get the desired username and password for our webserver VMs.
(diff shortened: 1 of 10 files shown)


lukehoban:
LGTM. 

The raw part of this is nearly identical to `azure-js-webserver`. Should we remove that one?

azure-ts-webserver-component/webserver.ts
```diff
@@ -0,0 +1,111 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as azure from "@pulumi/azure";
+import { specialArchiveSig } from "@pulumi/pulumi/runtime";
+
+// All resources will share a resource group.
+let resourceGroupName = new azure.core.ResourceGroup("server").name;
```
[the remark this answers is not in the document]
joeduffy: Good point. Will do.

azure-ts-webserver-component/index.ts
```diff
@@ -0,0 +1,21 @@
...
+export let ipAddresses = [];
+for (let i = 0; i < count; i++) {
+    let server = new WebServer(`ws-${i}`, {
+        username,
+        password,
+        bootScript: `#!/bin/bash\n
+echo "Hello, from Server #{i+1}!" > index.html
+nohup python -m SimpleHTTPServer 80 &`,
+    });
+    ipAddresses.push(server.getIpAddress());
```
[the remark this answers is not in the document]
joeduffy: `map` would certainly read nicer than a mutable array. It might be a little foreign, but then again, that's one of the benefits of using JavaScript here. @lukehoban @CyrusNajmabadi Any opinion on this?

joeduffy:
> The raw part of this is nearly identical to azure-js-webserver. Should we remove that one?

I don't mind having both, actually, to cater to peoples' different language needs. In general, I don't love maintaining multiple copies of the same thing, but there are real differences here.

azure-ts-webserver-component/index.ts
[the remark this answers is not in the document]
CyrusNajmabadi: i thnk the for-loop reads quite well.  So i wouldn't over-index here.  I find with heavyweight stuff (like making resources) sometimes being a bit more imperative/explicit feels a little better (to me at least).

azure-ts-webserver-component/webserver.ts
```diff
@@ -0,0 +1,111 @@
...
+     */
+    constructor(name: string, args: WebServerArgs) {
+        super("ws-ts-azure-comp:webserver:WebServer", name);
+
+        // Allocate a public IP and assign it to our NIC.
+        this.publicIp = new azure.network.PublicIp(`${name}-ip`, {
+            resourceGroupName,
+            allocationMethod: "Dynamic",
+        }, { parent: this });
+        this.networkInterface = new azure.network.NetworkInterface(`${name}-nic`, {
```
CyrusNajmabadi: we've been trying to move away from the pseudo hungarian `-xxx` suffixes for things.  I assume we're wildly inconsistent in examples.  But if we can prevent it from creeping in, we probably should.

azure-ts-webserver-component/webserver.ts
```diff
@@ -0,0 +1,111 @@
...
+    public readonly networkInterface: azure.network.NetworkInterface;
+    public readonly vm: azure.compute.VirtualMachine;
+
+    /**
+     * Allocate a new web server VM, NIC, and public IP address.
+     * @param name The name of the web server resource.
+     * @param args A bag of arguments to control the web server VM creation.
+     */
+    constructor(name: string, args: WebServerArgs) {
+        super("ws-ts-azure-comp:webserver:WebServer", name);
```
CyrusNajmabadi: seems unfortunate to not takes 'opts' here, just to be consistent with the Component pattern.  But it is an example... so maybe it matters less?

azure-ts-webserver-component/webserver.ts
```diff
@@ -0,0 +1,111 @@
...
+                version: "latest",
+            },
+        }, { parent: this });
+    }
+
+    public getIpAddress(): pulumi.Output<string> {
+        // The public IP address is not allocated until the VM is running, so wait for that
+        // resource to create, and then lookup the IP address again to report its public IP.
+        let ready = pulumi.all({
+            _: this.vm.id, name: this.publicIp.name, resourceGroupName: this.publicIp.resourceGroupName });
```
CyrusNajmabadi: cute.  consider making this a single line.

azure-ts-webserver/index.ts
```diff
@@ -0,0 +1,78 @@
...
+        offer: "UbuntuServer",
+        sku: "16.04-LTS",
+        version: "latest",
+    },
+});
+
+// The public IP address is not allocated until the VM is running, so wait for that
+// resource to create, and then lookup the IP address again to report its public IP.
+let done = pulumi.all({
+    _: vm.id, name: publicIp.name, resourceGroupName: publicIp.resourceGroupName });
```
CyrusNajmabadi: consider making a single line.

azure-ts-webserver/index.ts
```diff
@@ -0,0 +1,78 @@
...
+export const ipAddress = done.apply(d =>
+    azure.network.getPublicIP({
+        name: d.name, resourceGroupName: d.resourceGroupName }).then(ip => ip.ipAddress));
```
CyrusNajmabadi: consider async/await here.

CyrusNajmabadi:
Looking great to me.  Just minor suggestions.

azure-ts-webserver-component/webserver.ts
mikhailshilkov: This is actually [recommended by Microsoft](https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions):
> When naming resources in Azure, it is recommended to use common prefixes or suffixes to identify the type and context of the resource.
> Suggested Pattern
> [service short name]-[environment]-rg
> [vmname]-nic[num]

Any reason why you don't want to follow along?

azure-ts-webserver-component/index.ts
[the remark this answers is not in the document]
mikhailshilkov: I agree that `map` is more natural for pure functions, not ones creating cloud resources. But at the same time I keep writing snippets like this https://github.com/pulumi/pulumi-gcp/issues/158#issuecomment-498644482 and I would hate to change it to a for-loop. Might be a personal pref.

azure-ts-webserver-component/webserver.ts
CyrusNajmabadi: Apparently what's old is new again.  MS had hungarian, then moved away, and now moves back :)

If this is their recommendation, then this is totally fine.  

In general though, we've avoided this because it's generally redundant in pretty much any area we look in.  For example, if i'm on the AWS/Azure console then i'm enerally looking at resources that are already from some category.  i.e. i'm looking at my load-balancers.  So i don't need them to be suffixed with `-lb`.  Similarly, because a pulumi-urn already contains the type of the resource as part of it, it just ends up being a lot of redundancy.  i.e. if it already says `awsx:elb2:loadbalancer/...` i don't really need it to say `-lb` in the name to reaffirm that :)

But, like i said, if MS actively prefers this, then we should likely follow hte when-in-rome approach and accept that there.

mikhailshilkov: I don't have statistics, but it might be common to 
- look at a resource group with all sub-resources of different types
- search by name in the portal, and use resource type suffix while doing so

At least, I do so :)

azure-ts-webserver-component/webserver.ts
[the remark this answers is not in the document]
stack72: Ok, I have made these changes - will push to the branch

azure-ts-webserver-component/webserver.ts
stack72: I guess this is a noop now that we go back to the same approach

azure-ts-webserver-component/webserver.ts
stack72: done!

azure-ts-webserver/index.ts
stack72: done!

azure-ts-webserver-component/index.ts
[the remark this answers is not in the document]
stack72: We will leave this as is for now!

stack72:
The awesome @mikhailshilkov has agreed to make the amendment to async / await as per @CyrusNajmabadi's comment - we will merge this and then follow up on that

stack72:
Thanks for all the work here! This is passing it's CI tests :)

Issue #162, opened 2018-10-30
Update examples to not JSON.stringify policies


jen20:
Several of the examples still use JSON.stringify for policies instead of using the automatic transformers with literal PolicyDocument values. It would be better if the examples all used the automatic serialisation facilities, and if there was an example of unknown values being resolved correctly.

PR #310, opened 2019-06-05
Improve AWS Serverless RAW example

mikhailshilkov:
- Update to .NET Core 2.1 as 2.0 is not supported anymore
- Create policies with types rather than JSON.stringify (#162)
- Use a fine-grained policy to show how to set permissions

### DIFF
--- a/aws-ts-serverless-raw/index.ts
+++ b/aws-ts-serverless-raw/index.ts
@@ -29,27 +29,23 @@ let counterTable = new aws.dynamodb.Table("counterTable", {
 
 // Give our Lambda access to the Dynamo DB table, CloudWatch Logs and Metrics.
 const role = new aws.iam.Role("mylambda-role", {
-    assumeRolePolicy: aws.iam.assumeRolePolicyForPrincipal({ Service: "lambda.amazonaws.com" })
+    assumeRolePolicy: aws.iam.assumeRolePolicyForPrincipal({ Service: "lambda.amazonaws.com" }),
 });
 
-const policy = new aws.iam.Policy("mylambda-policy", {
-    policy: counterTable.arn.apply(arn => aws.iam.getPolicyDocument({
-        statements: [{
-            actions: ["dynamodb:UpdateItem", "dynamodb:PutItem", "dynamodb:GetItem", "dynamodb:DescribeTable"],
-            resources: [arn],
-            effect: "Allow",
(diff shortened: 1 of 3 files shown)


aws-ts-serverless-raw/index.ts
```diff
@@ -27,31 +27,44 @@ let counterTable = new aws.dynamodb.Table("counterTable", {
...
-            "Effect": "Allow",
-            "Sid": "",
-        },
-    ],
-};
-let role = new aws.iam.Role("mylambda-role", {
-    assumeRolePolicy: JSON.stringify(policy),
+// Give our Lambda access to the Dynamo DB table, CloudWatch Logs and Metrics.
+const role = new aws.iam.Role("mylambda-role", {
+    assumeRolePolicy: pulumi.output(
```
lukehoban: While we are changing this - we could simplify further with https://pulumi.io/reference/pkg/nodejs/pulumi/aws/iam/#assumeRolePolicyForPrincipal.

aws-ts-serverless-raw/index.ts
```diff
@@ -27,31 +27,44 @@ let counterTable = new aws.dynamodb.Table("counterTable", {
...
+                }],
+                effect: "Allow",
+            }],
+        })
+    ).apply(r => r.json),
 });
-let fullAccess = new aws.iam.RolePolicyAttachment("mylambda-access", {
+
+const policy = new aws.iam.Policy("mylambda-policy", {
+    policy: counterTable.arn.apply(arn => aws.iam.getPolicyDocument({
```
lukehoban: You should not need `getPolicyDocument`. That is inherited from the Terraform provider, but should be unneeded in Pulumi since you can write JSON more directly. 

In fact this whole thing could be:

`policy: pulumi.output({ ...})`

Where the object inside can embed `counterTable.arm`.

mikhailshilkov: `policy` is of type `string`, so somebody has to check types first and then convert it?

lukehoban: Ahh - that sounds like a bug - opened https://github.com/pulumi/pulumi-aws/issues/583.

But the reason I thought it did was actually that this could (and is in most of our other examples) be simplified to `new aws.iam.RolePolicy()` instead of `aws.iam.Policy + aws.iam.RolePolicyAttachment`.  The former is both shorter and does already support passing JSON directly.

PR #316, opened 2019-06-17
Update Azure Functions example to HttpEventSubscription

mikhailshilkov:
Update the example that creates an HTTP Azure Function to use a built-in `HttpEventSubscription` class instead of hand-made one, according to https://github.com/pulumi/pulumi-azure/issues/261 and https://blog.pulumi.com/serverless-as-simple-callbacks-with-pulumi-and-azure-function


mikhailshilkov:
The deleted file is also failing the current build with a weird error that I don't understand. Hopefully, this PR will fix the build.

PR #317, opened 2019-06-17
Azure API Management

mikhailshilkov:
An example Pulumi program that deploys an instance of Azure API Management with the following resources:

- API which is linked to an Azure Function App backend
- Operation and operation policy with URL rewrite and caching rules
- A product, a user, and a subscription to enable access to the API


stack72:
Hey @mikhailshilkov 

This LGTM! Thanks so much for sending this in - will be a nice example for those looking at taking advantage of API Management

Paul

PR #318, opened 2019-06-18
aws-ts-serverless-raw: Update Lambda Runtime Version

stack72:
The `DotnetCore2d0Runtime` has been deprecated so this needs to be
bumped to `DotnetCore2d1Runtime`

This stops the example running


stack72:
@lukehoban this is the app that requires us to have DotNet installed to build the application - I think we can revisit this part at another time as it's listed in the issue you linked me to

mikhailshilkov:
@stack72 How about https://github.com/pulumi/examples/pull/310 ?

lukehoban:
Ahh - yes - looks like we should merge #310 instead of this one.

lukehoban:
Closing this out as #310 has been merged now which fixes this.

stack72:
👍 from me!

PR #321, opened 2019-06-20
Use azure.Locations instead of plain strings (and fix some AKS errors)

mikhailshilkov:
As `azure.Locations` are released now, I changed all hard-coded strings into the typed ones.
Had to fix one AKS-related error to make those examples runnable.


clstokes:
LGTM

PR #327, opened 2019-06-25
Do not hard code AMI in examples

ellismg:


ellismg:
The build has been failing the past few nights because we are now running the examples in `us-west-2` in our new account.  We moved everything to `us-west-2` in order to have a single region that we have to keep track of instead of splitting stuff across multiple regions and managing limit increases in multiple places..

stack72:
@ellismg I <3 this ! Thank you!

PR #329, opened 2019-06-27
Azure Stream Analytics example

mikhailshilkov:
An example Pulumi program that deploys an Azure Stream Analytics job to transform data in an Event Hub.

### DIFF
--- a/azure-ts-stream-analytics/Pulumi.yaml
+++ b/azure-ts-stream-analytics/Pulumi.yaml
@@ -1,0 +1,10 @@
+name: azure-ts-stream-analytics
+runtime: nodejs
+description: Uses an Azure Stream Analytics job to transform data in an Event Hub.
+template:
+  config:
+    azure:environment:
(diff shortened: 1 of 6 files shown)


stack72:
Hey @mikhailshilkov 	

So I can see you listed the inputs, but what do I do with them? Where do I put them and how can I see the outputs?

P.

mikhailshilkov:
@stack72 You need to send events with an Event Hub client, e.g. [ServiceBusExplorer](https://github.com/paolosalvatori/ServiceBusExplorer). 

Once https://github.com/pulumi/pulumi-azure/pull/289 lands, we can easily add an HTTP callback function to send those events from within the stack, if you think that would help.

stack72:
@mikhailshilkov I think it would help people by adding a wee bit of info in the README about what they would still need to do with this sample application is all - I was confused at the start because I don't know these systems. The rest of the example looks good!

stack72:
This looks so much more usable now - that small command has helped a tonne - thanks so much! :)

PR #339, opened 2019-07-06
Enable aws-ts-ruby-on-rails example

stack72:
We have replaced an old dependency with a getAmi lookup
this means that we can run the example without the cross dependency


stack72:
Self merging - the green test is enough for enabling this example!

PR #343, opened 2019-07-10
Do not use "latest" GKE engine in examples.

pgavlin:
Using a fuzzy version as an input to `nodeVersion` can cause spurious
diffs, as the state returned by the provider will contain a specific
version for this field. Instead, fetch the latest version from GKE using
`getEngineVersions` and use that.

### DIFF
--- a/gcp-py-gke/__main__.py
+++ b/gcp-py-gke/__main__.py
@@ -1,6 +1,6 @@
 from pulumi import Config, export, get_project, get_stack, Output, ResourceOptions
 from pulumi_gcp.config import project, zone
-from pulumi_gcp.container import Cluster
+from pulumi_gcp.container import Cluster, get_engine_versions
 from pulumi_kubernetes import Provider
 from pulumi_kubernetes.apps.v1 import Deployment
 from pulumi_kubernetes.core.v1 import Service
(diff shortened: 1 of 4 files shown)


gcp-py-gke/__main__.py
```diff
@@ -19,11 +19,13 @@
 # password is the password for the admin user in the cluster.
 PASSWORD = config.get_secret('password') or RandomString("password", length=20, special=True).result
 
+engine_version = Output.from_input(get_engine_versions()).latest_master_version
```
lukehoban: This is pretty "strange" looking code to have to add to the top of many of our intro-to-pulumi examples.  I'd actually be inclined to just hard code a version and have to fix up our examples occasionally as these get updated - or pick this up from config and hard code the values into the test config?

I don't love hurting readability/usability of examples just to make testing easier.

pgavlin: I'd love to make this easier to read, but I do kind of prefer this approach to the hard-coded version approach. If I knew Python better, I would have avoided the `Output.from_input` bit.

That said, this is important content, so I'll defer to you if you feel strongly here.

pgavlin: Hard-coding into the config seems reasonable-ish, as we could also pull the latest version inside the test framework if we cared to.

pgavlin:
Merged to unblock some testing. I've filed https://github.com/pulumi/examples/issues/344 for follow-up.

PR #345, opened 2019-07-12
Adding a DigitalOcean example to CI

stack72:


digitalocean-ts-loadbalanced-droplets/index.ts
```diff
@@ -0,0 +1,40 @@
...
+const dropletCount = 3;
+const region = digitalocean.Regions.NYC3;
+
+const dropletTypeTag = new digitalocean.Tag("demo-app");
+const userData =
+    `#!/bin/bash
+  sudo apt-get update
+  sudo apt-get install -y nginx`;
+const droplets = [];
+for (let i = 0; i < dropletCount; i ++) {
```
justinvp: Nit

```suggestion
for (let i = 0; i < dropletCount; i++) {
```

digitalocean-ts-loadbalanced-droplets/package.json
```diff
@@ -0,0 +1,9 @@
+{
+    "name": "webserver-linode",
+    "version": "0.1.0",
+    "main": "index.js",
```
justinvp:
```suggestion
```

digitalocean-ts-loadbalanced-droplets/package.json
```diff
@@ -0,0 +1,9 @@
+{
+    "name": "webserver-linode",
+    "version": "0.1.0",
```
justinvp: Do we need the version?

digitalocean-ts-loadbalanced-droplets/package.json
```diff
@@ -0,0 +1,9 @@
+{
```
justinvp: Should we include a dev dependency on the node types like we do in our templates (to make it easier to work with Node APIs if someone forks and expands on this example)?

https://github.com/pulumi/templates/blob/a819c3df60524824ec3b0162dbb3ddc448fe7bfc/typescript/package.json#L3-L5

digitalocean-ts-loadbalanced-droplets/package.json
stack72: It's on the others so I kept it there - I am happy to remove it as a clean sweep from them all if you think it would be better?

digitalocean-ts-loadbalanced-droplets/package.json
stack72: I'm more than happy to go with your suggestion here but we need to do a sweep of all the examples if so - if you wanna open an issue about this and the version then you can assign it to me and I can look at doing it!

digitalocean-ts-loadbalanced-droplets/package.json
justinvp: OK, let's keep it.

digitalocean-ts-loadbalanced-droplets/package.json
justinvp: OK, if we don't do it for the other examples, not a huge priority to go update all the others. Fine with keeping things as-is.

PR #348, opened 2019-07-18
migrating(eks): add example of migrating node groups with zero downtime

metral:
Adding a condensed example of the `pulumi/eks` [e2e test](https://github.com/pulumi/pulumi-eks/pull/195).

This example is related to the tutorial & blog post:
- Tutorial PR: https://github.com/pulumi/docs/pull/1327
- Blog Post PR: https://github.com/pulumi/docs/pull/1328

Also, refactor `assertHttpResultWithRetry()` to accept HTTP headers.

### DIFF
--- a/aws-ts-eks-migrate-nodegroups/iam.ts
+++ b/aws-ts-eks-migrate-nodegroups/iam.ts
@@ -1,0 +1,51 @@
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+import * as iam from "./iam";
+
+const managedPolicyArns: string[] = [
+    "arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy",
+    "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
+    "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly",
+];
+
+// Creates a role and attaches the EKS worker node IAM managed policies
+export function createRole(name: string): aws.iam.Role {
+    const role = new aws.iam.Role(name, {
+        assumeRolePolicy: aws.iam.assumeRolePolicyForPrincipal({
+            Service: "ec2.amazonaws.com",
+        }),
+    });
+
+    let counter = 0;
+    for (const policy of managedPolicyArns) {
+        // Create RolePolicyAttachment without returning it.
+        const rpa = new aws.iam.RolePolicyAttachment(`${name}-policy-${counter++}`,
+            { policyArn: policy, role: role },
+        );
+    }
+
+    return role;
+}
+
+// Creates a collection of IAM roles.
+export function createRoles(name: string, quantity: number): aws.iam.Role[] {
+    const roles: aws.iam.Role[] = [];
+
+    for (let i = 0; i < quantity; i++) {
+        roles.push(iam.createRole(`${name}-role-${i}`));
+    }
+
+    return roles;
+}
+
+// Creates a collection of IAM instance profiles from the given roles.
+export function createInstanceProfiles(name: string, roles: aws.iam.Role[]): aws.iam.InstanceProfile[] {
+    const profiles: aws.iam.InstanceProfile[] = [];
+
+    for (let i = 0; i < roles.length; i++) {
+        const role = roles[i];
+        profiles.push(new aws.iam.InstanceProfile(`${name}-instanceProfile-${i}`, {role: role}));
+    }
+
+    return profiles;
+}
--- a/aws-ts-eks-migrate-nodegroups/index.ts
+++ b/aws-ts-eks-migrate-nodegroups/index.ts
@@ -1,0 +1,83 @@
+import * as awsx from "@pulumi/awsx"; import * as eks from "@pulumi/eks";
+import * as k8s from "@pulumi/kubernetes";
+import * as pulumi from "@pulumi/pulumi";
+import * as echoserver from "./echoserver";
+import * as iam from "./iam";
+import * as nginx from "./nginx";
+import * as utils from "./utils";
+
+const projectName = pulumi.getProject();
+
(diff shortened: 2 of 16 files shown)


aws-ts-eks-migrate-nodegroups/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    instanceRoles: roles,
+    enabledClusterLogTypes: ["api", "audit", "authenticator",
+        "controllerManager", "scheduler"],
+});
+export const kubeconfig = myCluster.kubeconfig;
+export const clusterName = myCluster.core.cluster.name;
+
+// Create a Standard node group of t2.medium workers.
+const ngStandard = utils.createNodeGroup(`${projectName}-ng-standard`, {
+    ami: "ami-03a55127c613349a7", // k8s v1.13.7 in us-west-2
```
stack72: we should add a call to `aws.getAmi` for this - this will mean we are able to be able to run our tests elsewhere easily - locally I use eu-west-1 as it's better latency for me

aws-ts-eks-migrate-nodegroups/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    instanceType: "t2.medium",
+    desiredCapacity: 3,
+    cluster: myCluster,
+    instanceProfile: instanceProfiles[0],
+});
+
+// Create a 2xlarge node group of t3.2xlarge workers with taints on the nodes
+// dedicated for the NGINX Ingress Controller.
+const ng2xlarge = utils.createNodeGroup(`${projectName}-ng-2xlarge`, {
+    ami: "ami-0355c210cb3f58aa2", // k8s v1.12.7 in us-west-2
```
stack72: same comment about `getAmi`

Makefile
```diff
@@ -4,6 +4,17 @@ all: only_build only_test
 
 ensure:
 	cd misc/test && dep ensure -v
+	
+	# Install aws-iam-authenticator
+	# See: https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html)
+	curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.12.7/2019-03-27/bin/linux/amd64/aws-iam-authenticator
```
stack72: It's important to point out here that these will install linux versions only rather than on macOS - maybe we can have a case based on us being on darwin?

Makefile
```diff
@@ -4,6 +4,17 @@ all: only_build only_test
...
+	chmod +x ./aws-iam-authenticator
+	sudo mv aws-iam-authenticator /usr/local/bin
+	
+	# Install kubectl
+	curl -LO https://storage.googleapis.com/kubernetes-release/release/$(shell curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
```
stack72: Same comment re-macOS

aws-ts-eks-migrate-nodegroups/iam.ts
```diff
@@ -0,0 +1,51 @@
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+import * as iam from "./iam";
+
+const managedPolicyArns: string[] = [
+    "arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy",
```
stack72: question on these - if these are know policies that we need to tell people to use, should we add these to `aws.iam.ManagedPolicies` so that users get auto-complete on the correct policies to use?

aws-ts-eks-migrate-nodegroups/index.ts
metral: There currently is no good, programmatic way of retrieving these AMIs from AWS. https://github.com/aws/containers-roadmap/issues/231 has served as a catch all to capture, and track this need from AWS.

aws-ts-eks-migrate-nodegroups/iam.ts
metral: We've previously discussed this topic and the short of it is that this approach ultimately allows users full control over the roles, and maps well to our desired state model. See these threads for more info:

- https://github.com/pulumi/pulumi-eks/pull/98#issuecomment-479978253
- https://github.com/pulumi/pulumi-eks/pull/98#issuecomment-480452516

Makefile
metral: Hmm, this is a good point. This change was primarily intended for Travis CI testing, similarly to how we setup the `pulumi-eks` travis: https://github.com/pulumi/pulumi-eks/blob/master/.travis.yml#L12-L25.

Given that `examples` is user-facing, I'm not sure how we generally address this need outside of the `prereqs` section we call out in our examples & tutorials, e.g: https://www.pulumi.com/docs/reference/tutorials/kubernetes/tutorial-eks/#prerequisites

metral: Update: I'll move these from the `Makefile` to the `.travis.yml` file instead. the `aws-iam-authenticator` is actually already being installed in our [current travis config](https://github.com/pulumi/examples/blob/master/.travis.yml#L35-L36) so i'll remove these steps from the Makefile.

metral: Done

metral:
> I think we need to link to the blog post so that users know how to run the example. It kinda breaks the other convention of the README having step by step instructions

You're right. This one is tricky given that the tutorial is the step-by-step instructions here per the unique scenario. I'll open a PR to link the tutorial steps here once the tutorial PR is merged.

stack72:
> > I think we need to link to the blog post so that users know how to run the example. It kinda breaks the other convention of the README having step by step instructions
> 
> You're right. This one is tricky given that the tutorial is the step-by-step instructions here per the unique scenario. I'll open a PR to link the tutorial steps here once the tutorial PR is merged.

Sounds good to me! :)

PR #354, opened 2019-07-31
Move examples more off of 'cloud' onto aws/awsx.

CyrusNajmabadi:


aws-ts-url-shortener-cache-http/index.ts
```diff
@@ -2,12 +2,9 @@
 
 import * as pulumi from "@pulumi/pulumi";
 import * as aws from "@pulumi/aws";
-import * as awsx from "@pulumi/awsx";
 import * as cloud from "@pulumi/cloud-aws";
```
lukehoban: We can drop cloud here now?

CyrusNajmabadi: not just yet.  let me ping you about this.

PR #363, opened 2019-08-15
Remove migrate nodegroup test

metral:


metral:
Merging this in even though we have a failing CI.

This is a safe deletion of a test, and doing so would help unblock the ongoing failures in CI.

PR #365, opened 2019-08-19
Fix azure appservice docker example

mikhailshilkov:
Get rid of the instable Django image and replace it with a simple node-based one


stack72:
This is awesome! Thanks for making this example more rock solid

PR #372, opened 2019-08-29
Added tests that allow us to store state in S3, Azure and GCP Storage

stack72:
This duplicates an AWS test and then stores the state for that duplication
in an S3 bucket. It does the same for a GCP and Azure tests as well


stack72:
Depends on #376 and https://github.com/pulumi/pulumi/pull/3169

stack72:
Closing this out - we know AWS works and those tests have been moved to https://github.com/pulumi/pulumi-aws/pull/737

GCP and Azure need extra EnvVars set that we need to investigate more

PR #377, opened 2019-09-04
Update any of the core providers to be a 1.0.0 dependency

stack72:


aws-py-stepfunctions/requirements.txt
```diff
@@ -1,2 +1,2 @@
-pulumi>=0.16.4
-pulumi-aws>=0.16.2
+pulumi>=1.0.0
+pulumi-aws>=1.0.0
```
lukehoban: Total nit - but should we take this opportunity to be more consistent on using `pulumi-aws` instead of `pulumi_aws`?

(I think that was the direction we planned on standardizing as part of https://github.com/pulumi/templates/pull/48 which never got merged).

gcp-py-gke/requirements.txt
```diff
@@ -1,4 +1,4 @@
-pulumi>=0.17.4
-pulumi_gcp>=0.18.2
-pulumi_kubernetes>=0.22.0
-pulumi-random>=0.5.3
+pulumi>=1.0.0
+pulumi_gcp>=1.0.0
+pulumi_kubernetes>=1.0.0
+pulumi-random>=1.0.0
```
lukehoban: This one even mixes the two styles in the same example! :-)

aws-py-stepfunctions/requirements.txt
ellismg: FWIW, I had been using `pulumi-aws>=1.0.0,<2.0.0` here, since that mirrors what the `^` constraint means in node-land (ignoring 2.0 and later).

lukehoban:
@stack72 Should we go ahead and merge this as-is for now?

stack72:
Yes, the failures are transcient

PR #379, opened 2019-09-04
Use the built-in container resource

mikhailshilkov:
The latest Terraform version introduced a resource for Cosmos DB SQL Container. Use it in the example.


stack72:
@mikhailshilkov can you rebase this and we can get it merged

PR #381, opened 2019-09-06
aws-py-webserver: Made the example region agnostic using ec2 get_ami

stack72:


stack72:
Self merging this as the updated test still works in the test run

PR #389, opened 2019-09-13
Add an example for deploying Miniflux on AWS

cnunciato:
Miniflux is an open-source RSS server and reader. This PR adds an example showing how to deploy it as a container on AWS Fargate using a PostgreSQL backend.


cnunciato:
Merging this now, as I've verified it works using the config values applied in the tests.

Issue #371, opened 2019-08-28
Do a pass over examples for consistency


joeduffy:
Now that we're rendering examples as tutorials in the docs website, we should do a pass over the examples for consistency. We're leaning towards leveraging Digital Ocean's guidelines -- https://www.digitalocean.com/community/tutorials/digitalocean-s-technical-writing-guidelines -- which will force us to ensure good titles, clean up the H2/H3s, and be better about listing prerequisites. As we make progress on this, we should regenerate the docs tutorials.

Taken from https://github.com/pulumi/docs/issues/1577#issuecomment-524687612:

- [x] Check in tutorial guidelines (similar to [DO's](https://www.digitalocean.com/community/tutorials/digitalocean-s-technical-writing-guidelines))
- [x] Tidy up the example titles/READMEs to conform to them and ensure they still work (with consistent casing)
- [ ] Reconcile duplication between existing tutorials and examples
- [x] Figure out what to do with tutorials that don't already exist as examples
- [x] Link to a decent example tutorial from the template file

mikhailshilkov:
Asked a related question in the docs repo: https://github.com/pulumi/docs/issues/1577#issuecomment-526077425

bermudezmt:
Let's continue the discussion here:

> @mikhailshilkov This is definitely the major difference that @bermudezmt and I have been discussing what to do about. I think we want the examples READMEs to be written in a more tutorial style, so that they read better in both places. It doesn't hurt to have more detail on GitHub too. But the challenge is that this is going to be a fair bit more effort and there is a lot of debt to pay down since none of the examples are written this way.

> @bermudezmt What are you thinking about this these days?

I do think we should create future examples with the idea that it's going to be transformed into a tutorial for the docs site. And I really like Digital Ocean's guidelines. 

I wonder if we can actually have a note for existing examples that have been turned into tutorials. It will be similar to our "derived" note from our package reference pages (https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/dnsimple/). This way we standardize for future examples and tutorials, but do not burden ourselves with redoing _all_ of the READMEs for the existing ones.

It might be good to add the tutorials that don't exist as examples to this repo.

mikhailshilkov:
[...]

PR #391, opened 2019-09-14
Add guidelines and README template for contributing to examples

bermudezmt:
This should hopefully standardize our tutorials.

Partially addresses #371

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,3 +67,5 @@ Each example should include a README to give the readers a good walkthrough. It 
 
 See our [example README template](example-readme-template.md.txt) for detailed explanations on each section.
 
+> The contribution guidelines have been authored in September 2019 and are subject to further refinements and tweaks. Examples prior to September 2019 do not necessarily conform to these guidelines.
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,69 @@
+# Contributing to Pulumi Examples
+
+Pulumi welcomes contributions from the community and is excited to continue building out its collection of examples and tutorials that showcase the power of modern infrastructure as code. This guide is intended to
+help your contribution get published quickly.
+
+## Pulumi terminology
+
+For a list of Pulumi terms used throughout our documentation site, see the [Glossary page](https://www.pulumi.com/docs/reference/glossary/).
+
+## Example folder structure
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+```
+|____README.md
+|____www
+|____images
+| |____deploy.gif
+| |____deploy.cast
+|____Pulumi.yaml
+|____package.json
+|____tsconfig.json
+|____index.ts
```
mikhailshilkov: Why do you have underscores in front of the file names?

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+- Overview paragraph
+- Prerequisites
+- Deploy the App
+   - Step 1 — Doing the First Thing
+   - Step 2 — Doing the Next Thing
+   …
+   - Step n — Doing the Last Thing
+- Clean Up
+- Summary
+- Next Steps (Optional)
```
mikhailshilkov: Should we explicitly admit that most of our existing samples don't really follow this structure and why?

example-readme-template.md.txt
```diff
@@ -0,0 +1,93 @@
...
+
+## Next Steps
+
+<!-- Optionally include an unordered list of relevant Pulumi tutorials. -->
+
+<!-- Example:
+- [Create a load-balanced, hosted NGINX container service](https://www.pulumi.com/docs/tutorials/aws/ecs-fargate/)
+- [Create an EC2-based WebServer and associated infrastructure](https://www.pulumi.com/docs/tutorials/aws/ec2-webserver/)
+-->
+
```
mikhailshilkov: Not sure where, but I would add a link to a reference README file that we think is the best example.

CONTRIBUTING.md
bermudezmt: This is the output of `tree`. I've always liked using it for showing a directory structure.

CONTRIBUTING.md
bermudezmt: Good call.

example-readme-template.md.txt
bermudezmt: I would do this as a follow-up change after editing one of the featured tutorials: Now tracked as an item in https://github.com/pulumi/examples/issues/371.

PR #401, opened 2019-09-20
Fix 1757 and clean up the tutorial

bermudezmt:
Fixes https://github.com/pulumi/docs/issues/1757


bermudezmt:
Thank  you for your suggestions, @mikhailshilkov !

PR #405, opened 2019-09-24
Add consistency parameters to work around the bug if TF

mikhailshilkov:
Workaround for https://github.com/terraform-providers/terraform-provider-azurerm/issues/4118, to be removed after we upgrade to 1.34


mikhailshilkov:
Self-merging as this mostly reverts the yesterday's "fix"

PR #418, opened 2019-10-10
Generate a random string for aws-ts-appsync datasource name

### LINKED ISSUE #420: Reenable aws-ts-appsync
There is an issue with the datasource name that we need to resolve - PR in progress here in #418

stack72:
Fixes #420

### DIFF
--- a/aws-ts-appsync/index.ts
+++ b/aws-ts-appsync/index.ts
@@ -1,6 +1,7 @@
 // Copyright 2016-2019, Pulumi Corporation.  All rights reserved.
 
 import * as aws from "@pulumi/aws";
+import * as random from "@pulumi/random";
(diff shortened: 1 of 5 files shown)


aws-ts-appsync/index.ts
```diff
@@ -43,8 +44,13 @@ const apiKey = new aws.appsync.ApiKey("key", {
     apiId: api.id,
 });
 
+const randomString = new random.RandomString("random-datasource-name", {
+    length: 15,
```
mikhailshilkov: You should limit this to `[_A-Za-z][_0-9A-Za-z]`, i.e. exclude special characters from `RandomString` options.

stack72: 👍 good call!

PR #422, opened 2019-10-11
Run examples tests in jobs, not stages.

pgavlin:
Jobs are considered to run concurrently: the failure of one job does not prevent other jobs from running. This is in contrast to stages, where the failure of some stage prevents later stages from running. In addition, jobs are allowed to run in parallel (though in this repo we have parallel builds disabled).

This also fixes the URL for the build failure notifier.


pgavlin:
I've verified that this does allow us to schedule all four jobs in such a way that the failure of one will not prevent the others from running. See e.g. https://travis-ci.com/pulumi/examples/builds/131609601

PR #423, opened 2019-10-14
Examples: Change the examples to run as individual tests

stack72:


stack72:
Self merging so I can disable some failing tests easily to make sure we are green

PR #425, opened 2019-10-14
Move examples from Pulumi-Policy repo

ekrengel:
Moving these examples from deep within the Pulumi-Policy repo to a subfolder here `/policy-packs`.

We will want to add some form of tests here in the near future.

### DIFF
--- a/policy-packs/README.md
+++ b/policy-packs/README.md
@@ -1,0 +1,5 @@
+# Example Policy Packs
+
+Status: Beta release.
+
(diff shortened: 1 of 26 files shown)


ekrengel:
> Convention wise, would it make more sense to just put examples in the /<pack-name>/ directory, rather than /policy-packs/<pack-name>/?

I dont think so. Policy s a subset of the product and all of the other examples are actual Pulumi projects. By putting them in their own sub-folder, we can have one nice README.md` explaining Policy Packs and how they fit into the Pulumi ecosystem. Else, I worry someone will stumble onto an example Policy Pack and be very confused.

PR #429, opened 2019-10-15
Temporarily disable Azure examples from tests due to errors

stack72:


misc/test/examples_test.go
```diff
@@ -370,7 +371,7 @@ func TestAccAzureJsWebserver(t *testing.T) {
 }
 
 func TestAccAzurePyAks(t *testing.T) {
-	t.Skip("Skipping due to ServicePrincipalNotFound error")
```
lukehoban: Do we want to keep some record of this other reason this won’t work to make it easier to reenable other azure tests?

lukehoban:
Test failure was unrelated to these changes - going ahead and merging.

PR #432, opened 2019-10-17
Adding an aws-stackreference-architecture example

stack72:


aws-stackreference-architecture/application/src/application.ts
```diff
@@ -0,0 +1,169 @@
...
+                                value: args.dbPort,
+                            },
+                            {
+                                name: "DB_NAME",
+                                value: args.dbName,
+                            },
+                        ],
+                    },
+                },
+                // tags: {
```
jen20: This seems like it should either be uncommented or removed?

jen20:
This LGTM with some minor nits around uncommented fields etc.

aws-stackreference-architecture/application/src/application.ts
stack72: Fixed!

PR #437, opened 2019-10-23
Add an example for running Cypress specs on a schedule

cnunciato:
This is still a work in progress, but here's an example of a Pulumi program that runs the [Cypress test runner](https://www.cypress.io/) headlessly on a five-minute schedule using the spec file(s) provided, capturing the test results (and optionally posting to Slack) on failure.


stack72:
Hey @cnunciato - going to close this out as it's very stale and hasn't had any movement in almost 18 months

PR #441, opened 2019-10-25
Update the examples to use the latest API proposal

justinvp:
Fixes https://github.com/pulumi/pulumi-policy/issues/133

### DIFF
--- a/policy-packs/aws-advanced/compute.ts
+++ b/policy-packs/aws-advanced/compute.ts
@@ -13,32 +13,39 @@
 // limitations under the License.
 
 import * as aws from "@pulumi/aws";
-import { Policy, typedRule } from "@pulumi/policy";
-import * as assert from "assert";
+import {
+    ReportViolation,
+    ResourceValidationArgs,
+    ResourceValidationPolicy,
+    validateTypedResource,
+} from "@pulumi/policy";
 
 export function requireApprovedAmisById(
     name: string,
     approvedAmis: string | Iterable<string>,
-): Policy {
+): ResourceValidationPolicy {
     const amis = toStringSet(approvedAmis);
 
     return {
         name: name,
-        description: "Instances should use approved AMIs",
+        description: "Instances should use approved AMIs.",
         enforcementLevel: "mandatory",
-        rules: [
-            typedRule(
-                aws.ec2.Instance.isInstance,
-                it => amis && assert.ok(amis.has(it.ami), "foo"),
-            ),
-            typedRule(
-                aws.ec2.LaunchConfiguration.isInstance,
-                it => amis && assert.ok(amis.has(it.imageId)),
-            ),
-            typedRule(
-                aws.ec2.LaunchTemplate.isInstance,
-                it => amis && assert.ok(it.imageId === undefined || amis.has(it.imageId)),
-            ),
+        validateResource: [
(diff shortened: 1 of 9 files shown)


policy-packs/aws-advanced/compute.ts
```diff
@@ -13,51 +13,60 @@
...
-            ),
-            typedRule(
-                aws.ec2.LaunchConfiguration.isInstance,
-                it => amis && assert.ok(amis.has(it.imageId)),
-            ),
-            typedRule(
-                aws.ec2.LaunchTemplate.isInstance,
-                it => amis && assert.ok(it.imageId === undefined || amis.has(it.imageId)),
-            ),
+        validateResource: [
```
justinvp: I know we discussed how it didn't make a lot of sense to have `rules` which accepted a single `rule`, and how the assumption is that most polices will have a single rule, but I found when updating the examples it was just easier to be able to provide multiple validation functions to be called for certain policies that are checking similar things across multiple resource types (like this case). Particularly when using the `validateTypedResource` helper to provide stronger typed resource objects to check. However, instead of making the property name plural, I've left it singular as the majority of policies will only have a single validation function -- but still allow multiple validation functions if it's easier to implement.

More discussion of this in https://github.com/pulumi/pulumi-policy/pull/131

policy-packs/aws-advanced/compute.ts
```diff
@@ -160,30 +191,34 @@ export function requireDetailedMonitoring(name: string): Policy {
...
-            assert.ok(it.encrypted);
-            if (kmsKeyId !== undefined) {
-                assert.strictEqual(it.kmsKeyId, kmsKeyId);
+        validateResource: validateTypedResource(aws.ebs.Volume.isInstance, (it, args, reportViolation) => {
+            if (!it.encrypted) {
+                reportViolation("EBS volumes should be encrypted.");
+            }
+            if (kmsKeyId !== undefined && it.kmsKeyId !== kmsKeyId) {
+                reportViolation(`EBS volumes should be encrypted with KMS ID '${kmsKeyId}'.`);
             }
```
justinvp: An example of where you might not want to necessarily return early (which happens with the assert case), but rather return as many violations as possible.

policy-packs/aws-advanced/compute.ts
```diff
@@ -160,30 +191,34 @@ export function requireDetailedMonitoring(name: string): Policy {
...
 // https://docs.aws.amazon.com/config/latest/developerguide/eip-attached.html
 
-export function requireEbsEncryption(name: string, kmsKeyId?: string): Policy {
+export function requireEbsEncryption(name: string, kmsKeyId?: string): ResourceValidationPolicy {
     return {
         name: name,
         description: "EBS volumes should be encrypted",
         enforcementLevel: "mandatory",
-        rules: typedRule(aws.ebs.Volume.isInstance, it => {
-            assert.ok(it.encrypted);
```
justinvp: It is slightly annoying always having to specify a message to `reportViolation` when previously we could somewhat get by with Node's `assert`'s provided message (however good/bad). Maybe it's a good thing to always force an explicit message about a violation? But I found that often the message I want to pass to `reportViolation` is the same or extremely similar to the policy's overall `description`.

justinvp:
> Although my guess is that we dont want examples pointing to dev for too long?

Right. We'll change it to an actual version soon once it's been released.

PR #444, opened 2019-10-27
Re-enable the Azure-py-aks example

stack72:


lukehoban:
This has failed 3 times in master since being reenabled. @stack72 let’s disable or make sure the test can be made reliable.

PR #445, opened 2019-10-27
Add .NET examples

mikhailshilkov:
The first set of examples in .NET


azure-cs-cosmosapp-component/CosmosApp.cs
```diff
@@ -0,0 +1,184 @@
...
+using System;
+using System.Collections.Immutable;
+using System.Linq;
+
+using Pulumi;
+using Pulumi.Azure.Core;
+using Pulumi.Azure.Cosmosdb;
+using Pulumi.Azure.Cosmosdb.Inputs;
+using Pulumi.Azure.Trafficmanager;
+using Pulumi.Azure.Trafficmanager.Inputs;
```
mikhailshilkov: I think it doesn't make sense to put child argument types into `Inputs` namespace. I end up importing it pretty much always, and until then the types aren't visible. cc @pgavlin

azure-cs-cosmosapp-component/VmScaleSets.cs
```diff
@@ -0,0 +1,293 @@
...
+        private ResourceGroup resourceGroup;
+
+        public Builder(ResourceGroup resourceGroup)
+        {
+            this.resourceGroup = resourceGroup;
+        }
+
+        public Func<RegionalContext, IRegionalEndpoint> BuildVMScaleSetApp(GlobalContext context)
+        {
+            var options = (CustomResourceOptions)ResourceOptions.Merge(context.Options, new CustomResourceOptions { DeleteBeforeReplace = true });
```
mikhailshilkov: This line feels a bit too tricky: I have to pass `new CustomResourceOption` (not `new ResourceOptions`) and then cast the result to `CustomResourceOption`. Maybe it should be `= CustomResourceOption.Merge(...same thing...)` and not accept `ResourceOptions`? @CyrusNajmabadi

azure-cs-webserver/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            var network = new VirtualNetwork("server-network",
+                new VirtualNetworkArgs
+                {
+                    ResourceGroupName = resourceGroup.Name,
+                    AddressSpaces = { "10.0.0.0/16" },
+                    Subnets =
+                    {
+                        new VirtualNetworkSubnetsArgs { Name = "default",  AddressPrefix = "10.0.1.0/24" }
+                    },
```
mikhailshilkov: This code yields a warning (as any other `InputList` does): `Object or collection initializer implicitly dereferences possibly null member 'VirtualNetworkArgs.Subnets'`.
Any reason why we mark those lists as nullable? @pgavlin @CyrusNajmabadi

azure-cs-webserver/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+                }, new CustomResourceOptions { DeleteBeforeReplace = true });
+
+
+            // The public IP address is not allocated until the VM is running, so wait for that
+            // resource to create, and then lookup the IP address again to report its public IP.
+            var ipAddress = Output
+                .Tuple<string, string, string>(vm.Id, publicIp.Name, resourceGroup.Name)
+                .Apply<string>(async t => {
+                    (_, string name, string resourceGroupName) = t;
+                    var ip = await Pulumi.Azure.Network.Invokes.GetPublicIP(new GetPublicIPArgs{ Name = name, ResourceGroupName = resourceGroupName });
```
mikhailshilkov: Because many namespaces have an `Invokes` class, I have to fully-qualify the namespace (or make a `using` alias). We may consider a different structure with a variable name for `Invokes`. @pgavlin

azure-cs-appservice/SharedAccessSignature.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Pulumi;
+using Pulumi.Azure.Storage;
+using Pulumi.Azure.Storage.Inputs;
+
+ public static class SharedAccessSignature
```
andreujuanc: Can he sneak in this class into the SDK? :)

mikhailshilkov: Yes, that's in the list https://github.com/pulumi/pulumi/issues/3406

azure-cs-cosmosapp-component/VmScaleSets.cs
andreujuanc: I don't like that cast as well. Would prefer to have something cleaner and faster to read.

azure-cs-cosmosapp-component/VmScaleSets.cs
```diff
@@ -0,0 +1,293 @@
...
+                        TargetResourceId = scaleSet.Id,
+                    },
+                    options);
+
+                return new AzureEndpoint(publicIp.Id);
+            };
+
+        }
+    }
+}
```
andreujuanc: File is quite big, maybe we can break it up further :)

azure-cs-appservice/SharedAccessSignature.cs
andreujuanc: Brilliant

azure-cs-cosmosapp-component/VmScaleSets.cs
mikhailshilkov: I'll keep the same structure as the TS one for now

PR #454, opened 2019-11-08
CSharp GKE example

clstokes:


clstokes:
@lukehoban @mikhailshilkov this is my first C# app ever, so let me know if anything isn't as idiomatic as it should be.

gcp-cs-gke/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+using Pulumi.Gcp.Container.Inputs;
+using Pulumi.Gcp.Container.Outputs;
+
+class Program
+{
+    static Task<int> Main()
+    {
+        return Deployment.RunAsync(async () =>
+        {
+
```
mikhailshilkov: Nit: remove the empty line

gcp-cs-gke/Program.cs
```diff
@@ -0,0 +1,82 @@
...
+            var name = "helloworld";
+
+            var config = new Config();
+
+            var latestMasterVersion = await Pulumi.Gcp.Container.Invokes.GetEngineVersions(new GetEngineVersionsArgs { });
```
mikhailshilkov: Nit: You don't need to fully quality `Invokes` in this case (until you import another namespace)

PR #455, opened 2019-11-11
Update PaC examples for latest API

justinvp:
Updates for the changes to no longer use the "output" shape of the resource (instead we're using the "input" shape, i.e. the resource's args type).

Depends on https://github.com/pulumi/pulumi-policy/pull/139


justinvp:
> We can just use `aws.s3.Bucket` because the compiler checks for an `isInstance` method on the type? And will call that should it exists?

At runtime, we check for the existence of an `isInstance` function, and call it if it is present.

PR #456, opened 2019-11-12
Update README.md to include the proper curl flag

prasannapad:
Currently, invoking the curl command to add the short URL fails to update CosmosDB with the indexed URL and gives error Failed to open {"id":"pulumi","url":"https://pulumi.com"}.

This PR addresses the issue - when making a curl request, pass in the correct -d flag for the POST body instead of -D.


mikhailshilkov:
Thank you @prasannapad

PR #460, opened 2019-11-15
Fix firewall rules not applied

### LINKED ISSUE #459: C# Firewall rules not applied
In this [example code](https://github.com/pulumi/examples/blob/ea36dc2d31508835485f3d9b45cc9ae7fc5bf9f2/azure-cs-msi-keyvault-rbac/Program.cs#L183) the firewall rules are not applied. Adding a `ToList()` after the `Select` function would apply the firewall rules.

alaatm:
Fixes #459


mikhailshilkov:
Thank you @alaatm!

PR #472, opened 2019-11-20
Fixing up the digitalocean-ts-k8s droplet slug

stack72:


stack72:
@pgavlin i was kinda going with the example set in Cyrus’s document from before. This doesn’t feel like a breaking change as it doesn’t cause a recreation of any resources. I agree it’s not ideal though but when we automate these, we are doing to fall into the same category where an instance size isn’t supported and therefore we won’t have it in our union helpers

Totally up for discussing the best way forward in these though as I’m not tied to a specific direction

pgavlin:
That seems reasonable enough to me. We'll need to do a release of the DO package in concert with this PR in order to keep things green.

PR #475, opened 2019-11-21
Add tests for C# and F#

lukehoban:


lukehoban:
Note that I don't think our integration test framework correctly supports the matrixed "latest" tests yet for .NET - would be great to add support there so that our nightly coverage of these tests gives us testing against head of packages and CLI updates.

PR #482, opened 2019-11-26
Example with docker and ACI for C# and F#

mikhailshilkov:
Two examples that show `docker.Image` usage for .NET SDK

Depends on https://github.com/pulumi/pulumi-docker/pull/115 being available in NuGet


azure-cs-cosmosapp-component/Azure.CosmosAppComponent.csproj
```diff
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Pulumi.Azure" Version="1.4.0-preview" />
+    <PackageReference Include="Pulumi.Docker" Version="1.0.0" />
```
mikhailshilkov: This is a fake version that I tested locally with. It needs to change before we merge.

stack72:
@mikhailshilkov please can you rebase this when you get a chance :)

azure-cs-cosmosapp-component/Azure.CosmosAppComponent.csproj
stack72: @mikhailshilkov can you also update to the correct version of the Pulumi.Docker plugin as well :)

mikhailshilkov:
@stack72 Done, sorry it took me so long

PR #486, opened 2019-11-28
[WIP] Example of using CDK constructs from Pulumi programs

lukehoban:
This example highlights how you can use CDK constructs from within Pulumi applications.  CDK
Constructs can be mixed with other Pulumi resources, including both Pulumi AWS resources, as well as
Azure, GCP, Kubernetes and more.

This is currently very simple.  It can be extended in a few ways:
- [ ] Show details of individual resources within the stack during updates
- [ ] Manage `cdk.out` inside the component state
- [ ] Show off mixing-and-matching Pulumi AWS resources with CDK AWS resources
- [ ] Support "container assets"
- [ ] Get outputs from CDK constructs/stack as Pulumi `Output`s


rshade:
@lukehoban since this has changed a bunch can we close it?

PR #487, opened 2019-11-28
pulumi.azure changed the AKS API

stack72:
Without this change, the AKS cluster will always try and recreate


azure-ts-aks-keda/cluster.ts
```diff
@@ -6,6 +6,7 @@ import * as k8s from "@pulumi/kubernetes";
 import * as pulumi from "@pulumi/pulumi";
 import * as random from "@pulumi/random";
 import * as tls from "@pulumi/tls";
+import * as config from "../azure-ts-aks-helm/config";
```
mikhailshilkov: This line looks wrong?

PR #495, opened 2019-12-10
Workaround pulumi/pulumi#3621

stack72:


aws-ts-pulumi-webhooks/index.ts
```diff
@@ -21,15 +21,15 @@ const stackConfig = {
 
 // Just logs information from an incoming webhook request.
 function logRequest(req: awsx.apigateway.Request) {
-    const webhookID = req.headers["pulumi-webhook-id"];
-    const webhookKind = req.headers["pulumi-webhook-kind"];
+    const webhookID = req.headers !== undefined ? req.headers["pulumi-webhook-id"] : "";
```
mikhailshilkov: A more idiomatic pattern might be `const v = req.headers && req.headers["name"];`

PR #500, opened 2019-12-14
GCP-TS-K8s-Ruby - Fix Import/DockerFile and Readme Updates

danmastrow:
Fix to make sure it compiles on first run without any user changes (fails on missing import)
Fix the minor version of the docker file (fails with mismatch between ruby 2.5.1 which is specified in the gem files and 2.5.7 which is what the docker file retrieves.)
Updated readme for users trying out Pulumi


stack72:
Thanks so much for the work here @danmastrow 

This LGTM!

PR #504, opened 2019-12-19
Set masterVersion from config

cbenjemaa:


stack72:
Closed via #580 

I added 1 more small piece to the work to allow a default value

PR #509, opened 2020-01-03
[prototype] Add a Python EKS example using multilang support

lukehoban:
This example uses the prototype multilanguage support in pulumi/pulumi#3711 to expose the EKS package in Python and enable a `aws-py-eks` example to be built.

This relies on light modifications to the EKS package in https://github.com/pulumi/pulumi-eks/pull/299.


justinvp:
Closing as no longer relevant.

PR #510, opened 2020-01-06
fix missing location in ResourceGroup

julson84:
Example does not work correct. Location needs to be specified for the ResourceGroup.


mikhailshilkov:
@julson84 Thanks so much for noticing the issues and for the PR.

We want to encourage Pulumi users to set Azure regions via the standard configuration value `azure:config`, which most (but not all) other examples do. So, I'll close this PR in favor of #511.

PR #513, opened 2020-01-14
Update Provisioner example with ability to trigger dynamic provider change

praneetloke:
This PR updates the provisioner example with a new property in the `ConnectionArgs` interface that allows a user to provide a value, which can trigger a replacement should the value change. An ideal source of values is a hash calculated from the file itself. As the file changes, the dynamic provider will re-run the related steps, i.e. copy/execute. Without this change, updating the user-defined conf or script that the provisioner copies will never be re-executed after the first time.


aws-ts-ec2-provisioners/index.ts
```diff
@@ -54,11 +55,12 @@ const server = new aws.ec2.Instance("server", {
...
     securityGroups: [ secgrp.name ],
 });
-const conn = {
+const conn: provisioners.ConnectionArgs = {
     host: server.publicIp,
     username: "ec2-user",
     privateKey,
-    privateKeyPassphrase,
+	privateKeyPassphrase,
+	changeToken: getFileHash("myapp.conf"),
```
lukehoban: Some indentation appears to be off here?

aws-ts-ec2-provisioners/provisioners/index.ts
```diff
@@ -15,7 +15,12 @@ export interface ConnectionArgs {
     username?: pulumi.Input<string>;
     password?: pulumi.Input<string>;
     privateKey?: pulumi.Input<string>;
-    privateKeyPassphrase?: pulumi.Input<string>;
+	privateKeyPassphrase?: pulumi.Input<string>;
+	/**
+     * A string value to control the replacement of the provisioner during each update.
+     * Provide a stable value if you do not need the provisioner to run each time you run an update.
+     */
+    changeToken: string;
```
lukehoban: I feel like this doesn't quite belong on the `ConnectionArgs` - and might make more sense as a property on `CopyFile` args and `RemoteExec` args?  In terraform, this is often called `keepers` - perhaps use that terminology (I've never found any terminology I particularly like for this).

Note that for the specific case of changing when the file contents change - it might be nice to actually have a first class `hash` property.  Doing just that might be sufficient here?

Though looking at your original example - perhaps the reason you did it the way you did was to trigger both of these to re-run - which would not be possible with my suggestion above due to the lack of https://github.com/pulumi/pulumi/issues/838.

Feel free to keep this as is - but perhaps consider `keepers` as the name?

As for doc comment here - the inconsistency is a little jarring - perhaps add comments on all properties? And either adopt the same `//` style, or update all existing doc comments to `/** */` style?

aws-ts-ec2-provisioners/util.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import * as crypto from "crypto";
+import * as fs from "fs";
+import * as path from "path";
+
+/**
+ * Calculates a hash for all of the files under the scripts directory.
+ */
+export function getFileHash(filename: string): string {
+	const data = fs.readFileSync(path.join(__dirname, filename), {encoding: "utf8"});
```
lukehoban: FWIW - I think the current interpretation of `src: ` is relative to cwd, not to the `__dirname`.

aws-ts-ec2-provisioners/index.ts
praneetloke: Yeah my editor was set to use Tabs for this repo for some reason. I have fixed it now.

aws-ts-ec2-provisioners/provisioners/index.ts
praneetloke: > I feel like this doesn't quite belong on the ConnectionArgs - and might make more sense as a property on CopyFile args and RemoteExec args?

Yeah I agree -- it was a little hacky to put it with the `ConnectionArgs`, to be honest. I moved it to be a property of those component resources, as well as the provisioner dynamic resource's "state" input as well.

> Though looking at your original example - perhaps the reason you did it the way you did was to trigger both of these to re-run

Yes, that's correct. The user can control which one they would like to re-run every time the file(s) content changes.

> Feel free to keep this as is - but perhaps consider keepers as the name?

I don't like the name `keepers`, so I left the name `changeToken`, but can change it if you insist on using `keepers`.

aws-ts-ec2-provisioners/util.ts
praneetloke: I am not sure I understand this comment?

praneetloke:
@lukehoban it seems I neglected this PR for quite a while. Could you please take another look when you get a chance? I updated it based on your feedback.

stack72:
@praneetloke is this something you can add to the python provisioners as well? :)

praneetloke:
I will go ahead and merge this. The failure is unrelated to the changes in this sample. It was a GCP test failure. All other tests passed.

For posterity, here's the test that failed:
```
--- FAIL: TestAccGcpJsWebserver (83.35s)
##[error]    program.go:769: [/home/runner/.pulumi/bin/pulumi up --non-interactive --yes --skip-preview --event-log /tmp/p-it-fv-az115-gcp-js-web-fd99589c-events.json --tracing https://tracing.pulumi-engineering.com/collector/api/v1/spans] failed: exit status 255; retrying...
##[error]    program.go:769: [/home/runner/.pulumi/bin/pulumi up --non-interactive --yes --skip-preview --event-log /tmp/p-it-fv-az115-gcp-js-web-fd99589c-events.json --tracing https://tracing.pulumi-engineering.com/collector/api/v1/spans] failed: exit status 255; retrying...
    program.go:604: 
        	Error Trace:	program.go:604
        	            				examples_test.go:1175
        	Error:      	Received unexpected error:
        	            	[/home/runner/.pulumi/bin/pulumi up --non-interactive --yes --skip-preview --event-log /tmp/p-it-fv-az115-gcp-js-web-fd99589c-events.json --tracing https://tracing.pulumi-engineering.com/collector/api/v1/spans] did not succeed after 3 tries
```

praneetloke:
> is this something you can add to the python provisioners as well?

I'll try to add something there as well, but likely won't get to it soon. Is that ok?

PR #518, opened 2020-01-21
pulumi serverless data warehouse example

EvanBoyle:
I turned my demo into an example project. This will be paired with a blog post being published later this week. Further detail on implementation, what the library does, etc are in the README.


EvanBoyle:
We've got a clean build now. 

I'm going to go ahead and merge this so we can get the blog post staged. I'm happy to make some updates after the fact if someone has the chance to look through this!

PR #527, opened 2020-01-27
Update Go examples for v1.9.1.

pgavlin:


pgavlin:
Note that this won't really be mergeable until the latest pulumi-terraform-bridge is rolled out to pulumi-aws and a release of that repo has been done.

webframp:
I'm really curious about the change to dep and Gopkg.toml here instead of using go.mod when most projects are going the other way. Is this a temporary thing or what's the benefit/reasoning behind that? 

I'm hoping to use pulumi aws SDKs with [gopls](https://github.com/golang/tools/tree/master/gopls) and this might impact how I approach that. I'm also thinking about how to possibly migrate a team already using Go and cloudformation to use Pulumi.

EvanBoyle:
@webframp Upgrading our templates and examples to go.mod is on my TODO list. It should be done by the end of the month. Tracking issue: https://github.com/pulumi/pulumi/issues/3817

webframp:
Right on thanks @EvanBoyle! I’ll watch that issue and blast Ytse Jam in your honor when it’s merged.

PR #542, opened 2020-02-06
Port two Azure examples to Go.

pgavlin:
And fix the Gopkg.toml files for AWS examples.

### DIFF
--- a/azure-go-webserver-component/Pulumi.yaml
+++ b/azure-go-webserver-component/Pulumi.yaml
@@ -1,0 +1,13 @@
+name: ws-go-azure-comp
+runtime: go
+description: Basic example of an Azure web server accessible over HTTP
+template:
+  config:
+    azure:location:
+      description: The Azure location to use
+      default: westus
+    username:
+      description: The username used to configure the Virtual Machine
+    password:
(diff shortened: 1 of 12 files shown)


azure-go-webserver-component/Pulumi.yaml
```diff
@@ -0,0 +1,13 @@
...
+template:
+  config:
+    azure:location:
+      description: The Azure location to use
+      default: westus
+    username:
+      description: The username used to configure the Virtual Machine
+    password:
+      description: The password used to configure the Virtual Machine
+      secret: true
```
EvanBoyle: Do you think we should remove this until we have secret support finished?

pgavlin: No, these should work as advertised: we allowed secret (i.e. encrypted) config before we had any SDK support for secret values.

pgavlin:
CI failure is unrelated:

```
[ amples/aws-ts-ec2-provisioners ]   pulumi-nodejs:dynamic:Resource (config-provisioner):
4208[ amples/aws-ts-ec2-provisioners ]     error: Timed out while waiting for handshake
```

Issue #544, opened 2020-02-08
aws-stackreference-architecture/networking broken?


viq:
I'm trying to stand this up (in `eu-central-1`), and getting an error:
```
$ pulumi config
KEY          VALUE
aws:profile  sky
aws:region   eu-central-1
azCount      2
```
```
$ pulumi up
Previewing update (dev):

     Type                 Name                  Plan       Info
 +   pulumi:pulumi:Stack  networking-layet-dev  create     1 error
 
Diagnostics:
  pulumi:pulumi:Stack (networking-layet-dev):
    error: Running program '/home/user/pulumi-examples/aws-stackreference-architecture/networking' failed with an unhandled exception:
    TypeError: Cannot read property 'slice' of undefined
        at /home/user/pulumi-examples/aws-stackreference-architecture/networking/src/index.ts:23:56
        at Generator.next (<anonymous>)
        at /home/user/pulumi-examples/aws-stackreference-architecture/networking/src/index.ts:8:71
        at new Promise (<anonymous>)
        at __awaiter (/home/user/pulumi-examples/aws-stackreference-architecture/networking/src/index.ts:4:12)
        at main (/home/user/pulumi-examples/aws-stackreference-architecture/networking/src/index.ts:16:12)
        at Object.<anonymous> (/home/user/pulumi-examples/aws-stackreference-architecture/networking/src/index.ts:69:18)
        at Module._compile (internal/modules/cjs/loader.js:778:30)
        at Module.m._compile (/home/user/pulumi-examples/aws-stackreference-architecture/networking/node_modules/ts-node/src/index.ts:536:23)
        at Module._extensions..js (internal/modules/cjs/loader.js:789:10)
 
```

It sounds like maybe it's not able to get the availability zones? But I'm able to from cli:
```
$ aws --profile sky ec2 describe-availability-zones
{
    "AvailabilityZones": [
        {
            "State": "available",
            "Messages": [],
            "RegionName": "eu-central-1",
            "ZoneName": "eu-central-1a",
            "ZoneId": "euc1-az2"
        },
        {
            "State": "available",
            "Messages": [],
            "RegionName": "eu-central-1",
            "ZoneName": "eu-central-1b",
            "ZoneId": "euc1-az3"
        },
        {
            "State": "available",
            "Messages": [],
            "RegionName": "eu-central-1",
            "ZoneName": "eu-central-1c",
            "ZoneId": "euc1-az1"
        }
    ]
}
```
[...]

PR #547, opened 2020-02-10
Fixup StackReference Architecture example

stack72:
Fixes: #544


stack72:
Build fails due to:

`Error: Cannot find module 'js-yaml'`

aws-stackreference-architecture/networking/test.json
```diff
@@ -0,0 +1,951 @@
+{
```
mikhailshilkov: Is this file needed?

stack72: It's removed :) Good catch!

PR #549, opened 2020-02-11
Add aws-py-eks stack

kevanpeters:
Created example of `aws-py-eks` stack with all the required AWS extras. 

I see there is a PR open #509 using multilang support creating a similar example, I can rename this stack if needed as this does not use that tooling.


aws-py-eks/vpc.py
```diff
@@ -0,0 +1,104 @@
+from pulumi_aws import ec2
+
+REGION = "us-east-2"
```
stack72: We shouldn't hardcode this value

aws-py-eks/vpc.py
```diff
@@ -0,0 +1,104 @@
...
+    'vpc-ig',
+    vpc_id=vpc.id,
+    tags={
+        'Name' : 'pulumi-vpc-ig'
+    }
+)
+
+## Subnets
+
+vpc_subnet_1 = ec2.Subnet(
```
stack72: based on aws.get_availability_zones, we should be able to loop and create the subnets for each of the zones, rather than hardcoding

aws-py-eks/vpc.py
```diff
@@ -0,0 +1,104 @@
...
+            'cidr_block' : '0.0.0.0/0',
+            'gateway_id' : igw.id
+        }
+    ],
+    tags={
+        'Name' : 'pulumi-vpc-rt'
+    }
+)
+
+ec2.RouteTableAssociation(
```
stack72: This can be wrapped inside the loop with the availability zones

aws-py-eks/vpc.py
kevanpeters: This is now pulling region from the config

aws-py-eks/vpc.py
kevanpeters: oh cool, I was going to use a loop but wanted to be explicit. This is exactly what I needed thanks for pointing the get_availability_zones out!

kevanpeters:
@stack72 Sorry for the delay on this, I was on vacation when this was reviewed. 

I incorporated the changes you called out, I am now looping through the subnets and am also no longer hard coding the region but getting it from the config. (I hope this works for tests and is a correct pattern) 

I also cleaned up some of the formatting and style to be more inline with recent additions to this repo

PR #552, opened 2020-02-12
Adding aws-py-fargate example

stack72:


clstokes:
LGTM! Nice and simple.

Issue #442, opened 2019-10-27
Add more Go language examples


stack72:
We want to add more examples that show off the Go support we have

EvanBoyle:
These should run in CI (at least some subset).

ekrengel:
@EvanBoyle Please add some examples that you would like @tasiah to work on. We will start her off with this to get up to speed with the Go programming model.

EvanBoyle:
Here are some ideas for things that we could convert to go:

1. [aws-cs-lambda](https://github.com/pulumi/examples/tree/master/aws-cs-lambda)
2. [gcp-cs-function](https://github.com/pulumi/examples/tree/master/gcp-cs-functions)
3. [gcp-ts-gke](https://github.com/pulumi/examples/blob/master/gcp-ts-gke/index.ts) This is dependant on the kubernetes go work https://github.com/pulumi/pulumi-kubernetes/issues/70
4. [azure-cs-aks](https://github.com/pulumi/examples/tree/master/azure-cs-aks)

PR #553, opened 2020-02-12
Created example for creating an AWS Lambda in Go

tasiah:
Added an example for `aws-go-lambda` for #442.


aws-go-lambda/Makefile
```diff
@@ -0,0 +1,4 @@
+build::
+	GOOS=linux GOARCH=amd64 go build -o ./handler/handler ./handler/handler.go
+	zip -j ./handler/handler.zip ./handler/handler
+	go build -o go-lambda main.go
```
lukehoban: Do we need this last line?

aws-go-lambda/Makefile
```diff
@@ -0,0 +1,4 @@
+build::
```
lukehoban: It may be inevitable - but having a Makefile here means this may not work well on Windows. There are some notes on doing this in a cross platform compatible way in https://github.com/aws/aws-lambda-go/blob/master/README.md#for-developers-on-windows.

aws-go-lambda/README.md
```diff
@@ -0,0 +1,61 @@
...
+3. [Clone aws-go-lambda](https://github.com/aws/aws-lambda-go)
+
+### Steps
+
+After cloning this repo, run these commands from the working directory:
+
+1. Build the go files:
+
+```bash
+make build
```
lukehoban: Btw - I’m hopeful that once we fix https://github.com/pulumi/pulumi/issues/3593 we can automate this directly within the Pulumi program so that the separate step is not needed - but today I think it is so I like having this example as is until then.

joeduffy:
Wow, nice! 🎉

aws-go-lambda/Makefile
tasiah: In the [example](https://github.com/pulumi/examples/tree/master/aws-go-webserver#steps) I was referring to, the first step was to build go, but to be honest, I've never tried _not_ building it.

lukehoban: > EDIT: Just tried it on a different example, and seems like the changes to main.go don't persist unless I build it first.

My understanding is that if you have the binary there, then it will be used, but if you don't, then we will effectively `go run main.go`.  So what you are seeing might happen if you still have the binary there from a previous `go build`?

@EvanBoyle Is that right?  And if so, is it documented yet?

tasiah: Just tested, and you're right that pulumi builds for us. I'll make this change with #554.

PR #554, opened 2020-02-13
Added gcp-go-functions-raw

tasiah:
Added an example for GCP Functions with Go for #442


gcp-go-functions/README.md
```diff
@@ -0,0 +1,52 @@
...
+1. [Configure Pulumi for GCP](https://www.pulumi.com/docs/intro/cloud-providers/gcp/setup/)
+    - [Set up a Service Account](https://www.pulumi.com/docs/intro/cloud-providers/gcp/service-account/)
+
+### Steps
+
+1.  Build `main.go`:
+
+    ```bash
+    go build -o gcp-go-functions main.go
+    ```
```
lukehoban: I believe this step is not needed - right @EvanBoyle ?  I think if it's not necessary we should skip including it by default.

tasiah: Apparently so: https://github.com/pulumi/examples/pull/553#discussion_r379110361. I'll delete this step.

PR #563, opened 2020-02-14
Added gcp-go-functions

tasiah:
This example is very similar to #554 except with deployment code written in Go. For #442.

### DIFF
--- a/gcp-go-functions-raw/Gopkg.toml
+++ b/gcp-go-functions-raw/Gopkg.toml
@@ -30,8 +30,8 @@
   version = "1.9.1"
 
 [[constraint]]
-  name = "github.com/pulumi/pulumi-aws"
-  version = "1.20.0"
+  name = "github.com/pulumi/pulumi-gcp"
+  version = "2.6.0"
 
 [prune]
   go-tests = true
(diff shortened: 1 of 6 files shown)


stack72:
Hey @tasiah 

This looks ace! There are 2 missing things from this example (and 1 of them is missing in #554)

* `dep ensure` to satisfy the dependencies
* `go build .` before the pulumi up

Apart from that, I just took it for a test drive and it looks great!

```

```

lukehoban:
> `go build .` before the pulumi up

This is not needed (anymore) as of https://github.com/pulumi/pulumi/pull/3503, and should no longer be our default guidance for users.  I do not believe docs have been updated yet to reflect this.

stack72:
👍 ooohhh that's awesome

stack72:
@tasiah this is still missing the entry in the README for `dep update`

P.

tasiah:
@stack72 I tested it locally and it worked as expected even though I never ran `dep ensure`. Should I still include it?

stack72:
ah interesting... let's leave it for now - I guess we will need to update the other Go READMEs to remove it if it's in there :)

lukehoban:
`dep ensure` will definitely be needed in the general case. Without it you will pick up whatever happens to be in your GOPATH (and will have to `go get` it there if you haven’t yet) and references wont be versioned. 

Would definitely encourage adding this consistently to examples for now as it is necessary to use these from a clean slate.

Issue #556, opened 2020-02-14
Update azure-go examples


tasiah:
Both azure-go examples ([`azure-go-appservice`](https://github.com/pulumi/examples/tree/master/azure-go-appservice) and [`azure-go-webserver-component`](https://github.com/pulumi/examples/tree/master/azure-go-webserver-component)) need to be updated.

PR #564, opened 2020-02-17
Update azure-go-appservice

tasiah:
This just updates `azure-go-appservice` for #556. Note that when I try to run `pulumi up` locally, I actually get an error, which I'm pretty sure is just an issue with my GCP subscription:

```
azure:appservice:Plan (azure-apps-asp):
    error: Error creating/updating App Service Plan "azure-apps-asp48033c21" (Resource Group "azure-apps-rgafeb114a"): web.AppServicePlansClient#CreateOrUpdate: Failure sending request: StatusCode=401 -- Original Error: Code="Unauthorized" Message="The scale operation is not allowed for this subscription in this region. Try selecting different region or scale option." Details=[{"Message":"The scale operation is not allowed for this subscription in this region. Try selecting different region or scale option."},{"Code":"Unauthorized"},{"ErrorEntity":{"Code":"Unauthorized","ExtendedCode":"52020","Message":"The scale operation is not allowed for this subscription in this region. Try selecting different region or scale option.","MessageTemplate":"The scale operation is not allowed for this subscription in this region. Try selecting different region or scale option.","Parameters":["default"]}}]
```

If there is an easy fix for `"The scale operation is not allowed for this subscription in this region. Try selecting different region or scale option."`, please let me know for future reference.

### DIFF
--- a/azure-go-appservice/Pulumi.yaml
+++ b/azure-go-appservice/Pulumi.yaml
@@ -1,4 +1,4 @@
-name: azure-appservice
+name: azure-go-appservice
 runtime: go
 description: Creates Azure App Service with SQL Database and Application Insights
 template:
--- a/azure-go-appservice/README.md
+++ b/azure-go-appservice/README.md
@@ -33,6 +33,12 @@ with App Service.
     pulumi config set --secret sqlPassword <value>
     ```
 
+1. Restore your Go dependencies. This example currently uses [Dep](https://github.com/golang/dep) to do so:
+
+    ```bash
+    $ dep ensure
+    ```
+
 1.  Run `pulumi up` to preview and deploy changes:
 
     ``` 
--- a/azure-go-appservice/main.go
+++ b/azure-go-appservice/main.go
@@ -42,7 +42,7 @@ func main() {
 			ResourceGroupName: resourceGroup.Name,
 			Location:          resourceGroup.Location,
 			Kind:              pulumi.String("App"),
-			Sku: appservice.PlanSkuInputs{
+			Sku: appservice.PlanSkuArgs{
 				Tier: pulumi.String("Basic"),
 				Size: pulumi.String("B1"),
 			},
@@ -115,8 +115,8 @@ func main() {
 				"ApplicationInsights:InstrumentationKey": appInsights.InstrumentationKey,
 				"APPINSIGHTS_INSTRUMENTATIONKEY":         appInsights.InstrumentationKey,
 			},
-			ConnectionStrings: appservice.AppServiceConnectionStringsArray{
-				appservice.AppServiceConnectionStringsInputs{
+			ConnectionStrings: appservice.AppServiceConnectionStringArray{
+				appservice.AppServiceConnectionStringArgs{
 					Name: pulumi.String("db"),
 					Value: pulumi.Sprintf("Server=tcp:%s.database.windows.net;initial catalog=%s;user ID=%s;password=%s;Min Pool Size=0;Max Pool Size=30;Persist Security Info=true;",
 						sqlServer.Name, database.Name, username, passwd),
@@ -147,12 +147,12 @@ func signedBlobReadURL(ctx *pulumi.Context, blob *storage.ZipBlob, account *stor
 			containerName := args[2].(string)
 			blobName := args[3].(string)
 
-			sas, err := storage.LookupAccountBlobContainerSAS(ctx, &storage.GetAccountBlobContainerSASArgs{
+			sas, err := storage.GetAccountBlobContainerSAS(ctx, &storage.GetAccountBlobContainerSASArgs{
 				ConnectionString: connectionString,
 				ContainerName:    containerName,
 				Start:            "2019-01-01",
 				Expiry:           signatureExpiration,
-				Permissions:      storage.GetAccountBlobContainerSASPermissionsArgs{Read: true},
+				Permissions:      storage.GetAccountBlobContainerSASPermissions{Read: true},
 			})
 			if err != nil {
 				return "", err


EvanBoyle:
It would be great to know that this works before merging. Were you able to try this in a few different regions to try and work around the error message? I can try and spin this up later today.

tasiah:
I tried `eastus` and `westus` with no luck. I can try some more later today. I wish there was an easy just to look up which regions we could use instead of randomly picking some though 😅.

lukehoban:
Note that we have similar examples for other languages - so as long as we're doing the same thing in the same regions - I would assume it should behave the same?

Also as a general note - we run many of our examples under test in CI.  That includes several AppService examples.  Would be good to add a few of the Go examples to CI to get validation that they work now, and continue to work going forward.  See https://github.com/pulumi/examples/blob/master/misc/test/examples_test.go.

lukehoban:
cc @mikhailshilkov for any help on getting AppService examples running correctly in our test subscriptions/regions.

tasiah:
@lukehoban Do we have a way to run `examples_test.go` locally? I couldn't find any documentation on how to and even tried looking through travis logs.

mikhailshilkov:
@tasiah Are you using a free Azure subscription? This should not happen with our Pulumi subscription.
Do we want to include this example to the test run in CI?

lukehoban:
> Do we have a way to run examples_test.go locally?

In principle yes - we can just run `go test` in that folder (or click “run test” in vs code). But these tests rely on environment variables to connect to appropriate cloud provider test accounts, and that part you would need to manually replicate locally if you want to target the same accounts.

tasiah:
@mikhailshilkov Yes, I'm using the Pulumi Subscription:
```
tasiahalim@Tasias-MBP examples % az account list
A few accounts are skipped as they don't have 'Enabled' state. Use '--all' to display them.
[
  {
    "cloudName": "AzureCloud",
    "id": "0282681f-7a9e-424b-80b2-96babd57a8a1",
    "isDefault": true,
    "name": "Pulumi Subscription",
    "state": "Enabled",
    "tenantId": "706143bc-e1d4-4593-aee2-c9dc60ab9be7",
    "user": {
      "name": "tasia@pulumi.com",
      "type": "user"
    }
  }
]
```

Perhaps something else is wrong with my Azure configuration.

tasiah:
@mikhailshilkov Yes, I'm using the Pulumi Subscription:
```
tasiahalim@Tasias-MBP examples % az account list --output table

A few accounts are skipped as they don't have 'Enabled' state. Use '--all' to display them.
Name                 CloudName    SubscriptionId                        State    IsDefault
-------------------  -----------  ------------------------------------  -------  -----------
Pulumi Subscription  AzureCloud   0*******-****-****-****-***********1  Enabled  True
```

mikhailshilkov:
> Yes, I'm using the Pulumi Subscription

This is surprising. We create a lot of App Service Plans in other languages and run in CI, they seem [good](https://travis-ci.com/pulumi/examples/jobs/288982353).

I get `package /Users/mikhailshilkov/examples/azure-go-appservice: import "/Users/mikhailshilkov/examples/azure-go-appservice": cannot import absolute path` when I try to `pulumi up` this. Is there an additional guidance on how this should be run?

Do you get the same error on the `azure-ts-appservice` examples?

tasiah:
I actually _don't_ get the same error for `azure-ts-appservice`. I was comparing the two and it seems like the only difference is that [`signedBlobReadUrl`](https://github.com/pulumi/pulumi-azure/blob/master/sdk/nodejs/storage/zMixins.ts#L33) is defined in `@pulumi/azure`, but we define [`signedBlobReadUrl`](https://github.com/pulumi/examples/blob/c09365b3010e792a5aa48afa12c509c1788fd1c1/azure-go-appservice/main.go#L136) ourselves for `azure-go-appservice`. The implementation for both however looks the same.

I've never encountered `package /Users/mikhailshilkov/examples/azure-go-appservice: import "/Users/mikhailshilkov/examples/azure-go-appservice": cannot import absolute path` and I'm not sure why it would be complaining about an absolute path considering we don't import any absolute paths. For other examples, I've been running `dep ensure` before `pulumi up`, so maybe try that?

EvanBoyle:
@mikhailshilkov not sure if this is relevant to your issue, but `dep ensure` will only work from within a src/ folder in your $GOPATH.

mikhailshilkov:
Thank you @tasiah @EvanBoyle, moving to $GOPATH + `dep ensure` made the trick. Should we add this to README?

The example is deployed like a charm for me (I tried `westeurope`) so I'm still confused about those scaling errors...

@tasiah Minor thing, but could you rename the project to `azure-go-appservice` in `Pulumi.yaml`? Otherwise, I got "conflicts" when deploying the Go example after the TS one.

tasiah:
I still get `"The scale operation is not allowed for this subscription in this region. Try selecting different region or scale option."` even after setting the location to `westeurope`. Since it's working for you, I'm guessing my local configuration is wrong somewhere.

EvanBoyle:
@tasiah what do you think about getting this merged? Seems like we have confidence that it works given @mikhailshilkov's testing.

mikhailshilkov:
Yes, it does work for me. We could also include this test into the test suite to run it as a part of the build, if that works for Go already.

tasiah:
I don't mind getting it merged and including a test for this (and some of the other Go examples) in a separate PR. At this point, I'm fine with accepting that something's wrong with my local azure configuration, although I would feel a little better if someone else besides Mikhail can confirm that it works (@EvanBoyle or @lukehoban maybe?).

ekrengel:
> I would feel a little better if someone else besides Mikhail can confirm that it works 

I think you should just go ahead and add these to the tests as confirmation!

tasiah:
According to @stack72, we can't actually test the go examples yet :( 

https://github.com/pulumi/examples/pull/579#issuecomment-591307085

stack72:
Correct - we don’t auto install the dependencies - I’m working on it atm

EvanBoyle:
@tasiah I just tried these changes and got the same error you originally listed. I imagine that our Azure subscriptions are probably configured in the same way.

I did notice that this example is missing a `dep ensure` step in the readme. Would you mind adding that? 

From a little reading, it sounds like the above error is just due to a misconfigured subscription. We can contact azure support to get this resolved, but given the fact that @mikhailshilkov can deploy this I think we can go ahead and ship this update to customers.

mikhailshilkov:
@EvanBoyle Now I'm even more puzzled because I use the same subscription. Do you mind filing a support ticket? Or I can do it if you point me to an existing resource group where this happened for you.

PR #571, opened 2020-02-21
build and release theia docker image

EvanBoyle:
This builds a [theia](https://theia-ide.org/) docker image every time we merge to master.

This allows us to have an image that we can link to in our gettings started documentation for users who would prefer to use docker and a local, web-based IDE, rather than installing the whole toolchain. 

Once we publish the image, you can run it via:

```sh
docker run -it --init -p 3000:3000 pulumi/theia:latest
```

`localhost:3000` will have a web-based IDE with terminal, full pulumi toolchain, and all of our examples. 

Screen Shot 2020-02-21 at 10 45 49 AM


Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM pulumi/pulumi	
+
```
lukehoban: Perhaps move these theia files to `misc/theia`?

docker.package.json
```diff
@@ -0,0 +1,43 @@
+{
+    "private": true,
+    "theia": {
+        "frontend": {
+            "config": {
+                "applicationName": "Theia JavaScript/TypeScript Example",
```
lukehoban: Should this be different?

docker.package.json
```diff
@@ -0,0 +1,43 @@
...
+    },
+    "resolutions": {
+        "vscode-languageserver-protocol": "3.15.0-next.9",
+        "vscode-languageserver-types": "3.15.0-next.5",
+        "**/vscode-json-languageserver/**/vscode-languageserver": "6.0.0-next.1"
+    },
+    "devDependencies": {
+        "@theia/cli": "latest"
+    }
+}
```
lukehoban: This and elsewhere -fix trailing new lines?

stack72:
Hey @EvanBoyle 

As per cleanup, I am going to close this out - it's gone stale. If you want to re-address this then we can definitely do that

P.

PR #573, opened 2020-02-22
Port aws-ts-static-website example to Python

leezen:


aws-py-static-website/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+1.  Create a new stack:
+
+    ```bash
+    $ pulumi stack init website-testing
+    ```
+
+1.  Set the AWS region:
+
+    ```
```
chrsmith: Nit: Since you put `bash` as the language type for the previous code block, we should be consistent here and the other snippets. (Though it won't actually _do_ anything, but in a perfect world it would lead to better syntax highlighting when rendered on GitHub.)

aws-py-static-website/README.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+This is caused by CloudFront confirming the ETag of the resource before applying any updates.
+ETag is essentially a "version", and AWS is rejecting any requests that are trying to update
+any version but the "latest".
+
+This error will occurr when the state of the ETag get out of sync between the Pulumi Service
+and AWS. (Which can happen when inspecting the CloudFront distribution in the AWS console.)
+
+This will get fixed in Pulumi soon, but for the time being you can find workaround steps in
```
chrsmith: We now have the ability to run a `pulumi refresh` to pickup any newer ETag values. So we should update the text to just say that instead of indirecting through a GitHub issue. (Would you mind updating the `README.md` file for the TypeScript version of this example to use the same text, too?)

aws-py-static-website/__main__.py
```diff
@@ -0,0 +1,166 @@
...
+    parts = domain.split('.')
+    if len(parts) < 2:
+        raise Exception(f'No TLD found on ${domain}')
+    if len(parts) == 2:
+        return "", domain
+    subdomain = parts[0]
+    parts.pop(0)
+    return subdomain, '.'.join(parts) + '.'
+
+stack_config = Config('static-website')
```
chrsmith: This source file would benefit a lot from having some comments. Personally having idiomatic doc comments on the functions would be great, but at the very least having comments to help organize the code would make this a little less intimidating.

For example, something like:

```python
# Read the stack's configuration data.
stack_config = Config('static-website')
...
```

chrsmith: I've never used the Python SDK before, but you usually want to omit the argument to `pulumi.Config`. It will default to the project name, which is better since it means that things will probably do what you want if you ever rename the project by editing `Pulumi.yaml`. (Otherwise your existing stacks would still work, but any attempt to standup new stacks with the newer project name wouldn't ever find the configuration values.)

aws-py-static-website/__main__.py
```diff
@@ -0,0 +1,166 @@
...
+
+content_bucket = pulumi_aws.s3.Bucket('contentBucket',
+    bucket=target_domain,
+    acl='public-read',
+    website={
+        'index_document': 'index.html',
+        'error_document': '404.html'
+    })
+
+def crawl_directory(content_dir, f):
```
chrsmith: I'd strongly suggest putting a doc comment on this function, since it isn't immediately obvious what `f` is used for.

aws-py-static-website/__main__.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+
+crawl_directory(web_contents_root_path, bucket_object_converter)
+
+logs_bucket = pulumi_aws.s3.Bucket('requestLogs', bucket=f'{target_domain}-logs', acl='private')
+
+TEN_MINUTES = 60 * 10
+
+# provision a certificate if the arn is not provided
+if (certificate_arn is None):
```
chrsmith: The `(` `)` are technically unnecessary, right? Would it be more idiomatic to omit them?

aws-py-static-website/__main__.py
```diff
@@ -0,0 +1,166 @@
...
+        opts=ResourceOptions(parent=content_bucket)
+    )
+
+crawl_directory(web_contents_root_path, bucket_object_converter)
+
+logs_bucket = pulumi_aws.s3.Bucket('requestLogs', bucket=f'{target_domain}-logs', acl='private')
+
+TEN_MINUTES = 60 * 10
+
+# provision a certificate if the arn is not provided
```
chrsmith: The hill I have chosen to die on is that [Comments Should be Sentences](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html). So here, and else where, should be complete sentences starting with a capital letter and ending with a period.

I believe I am the only one at the company who feels strongly about this, so feel free to take this feedback as any other matter of opinion.

aws-py-static-website/__main__.py
leezen: Yup! Thanks for catching that

aws-py-static-website/__main__.py
leezen: That's fine. I think for an example, you're right that we should have clearer comments.

leezen:
Thanks for the review!

PR #574, opened 2020-02-23
Add sample showing AKS + private ACR

sfmskywalker:
The sample further demonstrates building & pushing a .NET Core docker image to the registry, and then deploying a container of that image onto the cluster.


azure-cs-aks-private-container-registry/Program.cs
```diff
@@ -0,0 +1,216 @@
+﻿// Copyright 2016-2019, Pulumi Corporation.  All rights reserved.
```
mikhailshilkov: Nit: 2020 :)

azure-cs-aks-private-container-registry/Program.cs
```diff
@@ -0,0 +1,216 @@
...
+using Pulumi.Random;
+using Pulumi.Tls;
+using ContainerArgs = Pulumi.Kubernetes.Types.Inputs.Core.V1.ContainerArgs;
+using SecretArgs = Pulumi.Kubernetes.Types.Inputs.Core.V1.SecretArgs;
+using ServiceArgs = Pulumi.Kubernetes.Types.Inputs.Core.V1.ServiceArgs;
+
+class Program {
+	static Task<int> Main() {
+		return Deployment.RunAsync(() => {
+			var resourceGroup = new ResourceGroup("aks-rg");
```
mikhailshilkov: Could you please use the standard indentation of 4 spaces everywhere? Thanks!

azure-cs-aks-private-container-registry/Program.cs
```diff
@@ -0,0 +1,216 @@
...
+
+			var randomPassword = new RandomPassword("password", new RandomPasswordArgs {
```
mikhailshilkov: Put `{` on a new line everywhere, thank you!

azure-cs-aks-private-container-registry/Program.cs
```diff
@@ -0,0 +1,216 @@
...
+				Build = "./SampleApplication",
+				Registry = new ImageRegistry {
+					Server = registry.LoginServer,
+					Username = registry.AdminUsername,
+					Password = registry.AdminPassword
+				},
+				ImageName = imageName
+			}, new ComponentResourceOptions {
+				Provider = k8sProvider
+			});
```
mikhailshilkov: I suggest you creating `options` once and then passing the same object everywhere

azure-cs-aks-private-container-registry/Program.cs
```diff
@@ -0,0 +1,216 @@
...
+						Port = 80
+					}
+				}
+			}, new CustomResourceOptions {
+				Provider = k8sProvider
+			});
+
+			return new Dictionary<string, object?> {
+				["kubeconfig"] = cluster.KubeConfigRaw,
+				["dockercfg-secret-name"] = dockerCfgSecret.Metadata.Apply(x => x.Name),
```
mikhailshilkov: How would I use this output?

sfmskywalker: You would use it when creating a `Deployment` from another service project/stack when you want to push & pull a docker container image there from the registry created in this stack. This sample project uses it itself on line 184:

```csharp
ImagePullSecrets = new LocalObjectReferenceArgs {
	Name = dockerCfgSecretName
}
```

Depending on one's setup, probably the "horizontal" stack wouldn't be deploying any pods to the cluster - that would be done from "vertical" stacks - in which case the exported docker registry container secret comes handy.

Let me know if I should clarify this with a comment or something, or if you prefer I do something else.

azure-cs-aks-private-container-registry/Program.cs
sfmskywalker: Mysteriously enough, the editorconfig in the samples root caused Rider to apply different settings - removing editorconfig allowed Rider to stick to my system-defaults (which includes 4 spaces everywhere, BSD style braces, etc. Anyway, I will fix.

azure-cs-aks-private-container-registry/Program.cs
sfmskywalker: Makes sense. Except that some objects accept `ComponentResourceOptions`, whilst others accept `CustomResourceOptions` (the majority of cases).

I will create a `CustomResourceOptions` object once and pass it along where applicable, and leave the `ComponentResourceOptions` as-is, since it's the only occurrence.

azure-cs-aks-private-container-registry/Program.cs
mikhailshilkov: Maybe add a short comment to the README about this, thank you!

mikhailshilkov:
Another tiny remark: @sfmskywalker could you add your example to the root README https://github.com/pulumi/examples#c-1

mikhailshilkov:
LGTM, but GitHub is unhappy about the conflict in README. Could you resolve that one and then I'll merge.

sfmskywalker:
I see. I updated my fork with latest from master and resolved the conflict.

PR #577, opened 2020-02-25
Created AKS example for Go

tasiah:


azure-go-aks/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Steps
+
+1. Create a new stack:
+
+    ```bash
+    $ pulumi stack init dev
+    ```
+
+1.  Login to Azure CLI (you will be prompted to do this during deployment if you forget this step):
```
EvanBoyle: we need a step in here for `dep ensure`

tasiah:
Looks like the build is failing due to  `aws-ts-scheduled-function/index.ts`, which I didn't touch: https://travis-ci.com/pulumi/examples/builds/150375817#L933. Should I fix it, merge anyways, or...?

EvanBoyle:
We should fix it, but no need to block on getting this in.

PR #579, opened 2020-02-25
Add Go examples to examples_test.go

tasiah:


stack72:
Hey @tasiah 

We can't really add the Go examples *yet* as we don't run dep ensure by default. This could be a time sink for you

P.

tasiah:
That...would make sense. Thanks for pointing it out 😅. I just created this quick draft since I was curious how Travis would run it.

stack72:
Closed via #667

PR #581, opened 2020-02-29
clean up of python examples, update virtual environment instructions

spara:
updated python instructions for venv, also cleaned up markdown

### DIFF
--- a/aws-py-fargate/README.md
+++ b/aws-py-fargate/README.md
@@ -31,12 +31,12 @@ Next, to deploy the application and its infrastructure, follow these steps:
     $ pulumi config set aws:region us-east-1 # any valid AWS region will work
     ```
 
-3.  Create a Python virtualenv, activate it, and install dependencies:
+3. Create a Python virtualenv, activate it, and install dependencies:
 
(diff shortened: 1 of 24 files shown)


aws-py-fargate/README.md
```diff
@@ -121,26 +121,26 @@ Next, to deploy the application and its infrastructure, follow these steps:
 
     Notice that `pulumi up` redeploys just the parts of the application/infrastructure that you've edited.
 
-    ```
```
lukehoban: Do these snippets render “better” with `bash`?

azure-py-webserver/README.md
```diff
@@ -129,7 +129,7 @@ the virtual machine that we are going to create.
     -   ├─ azure:network:PublicIp          server-ip                    deleted     
     -   └─ azure:core:ResourceGroup        server                       deleted     
     
-    Resources:
+    s:
```
lukehoban: I think this one was unintentional?

spara: fixed

aws-py-fargate/README.md
spara: Not sure if it renders better, but it stops the markdown linter from complaining; and since the prompt is typically bash, I went with bash.

PR #584, opened 2020-03-02
PaC-based integration testing example

mikhailshilkov:
An example that applies PaC-based integration testing to provisioning an AWS EKS cluster.

I took the code of [Unit Testing Your Infrastructure with Node.js and Mocha](https://www.pulumi.com/blog/unit-testing-infrastructure-in-nodejs-and-mocha/) and re-created the tests based on pure PaC instead of Mocha/Chai.

Running:

```bash
PULUMI_EXPERIMENTAL=true pulumi up --policy-pack ./tests
```

An examples of failure:

```
Diagnostics:
  pulumi:pulumi:Stack (pac-ts-eks-dev):
    error: update failed
 
Policy Violations:
    [mandatory]  tests-pack v1  eks-test (pac-ts-eks-dev)
    EKS integration tests.
    EKS Cluster 'my-cluster-eksCluster-90d55b2' has unknown VPC
```

Some observations:

1. I use stack policy instead of resource policy, because:
1.1. I want to make sure that the cluster exists at all (see the first assertion).
1.2. I want to test the shape of *outputs*, not *args*. VPC ID is a good example: in a resource policy, `clusterArgs.vpcConfig.vpcId` is actually undefined in the arguments because we [don't assing it](https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/cluster.ts#L350). Inputs can be tested in mock-based tests; here I want to test the outputs.
1.3. The fix to https://github.com/pulumi/pulumi-policy/issues/199 is not available yet.

2. The UX is full of policy language, it says nothing about testing. We should decide whether we want to keep it that way, or whether we want to introduce a completely new lingua specifically for testing and use PaC only behind the scenes.

3. The test still needs to know about `pulumi.runtime.isDryRun` in case it's run with `pulumi preview` and the resources don't exist yet.

4. I had to use AWS SDK to retrieve the default VPC because [data sources are not supported](https://github.com/pulumi/pulumi-policy/issues/194).

### DIFF
--- a/testing/pac-ts-eks/tests/index.ts
+++ b/testing/pac-ts-eks/tests/index.ts
@@ -42,7 +42,7 @@ const stackPolicy: policy.StackValidationPolicy = {
     },
 }
 
-const tests = new policy.PolicyPack("aws-eks-unit-tests", {
+const tests = new policy.PolicyPack("tests-pack", {
     policies: [stackPolicy],
 });
 
--- a/testing/pac-ts-eks/tests/index.ts
+++ b/testing/pac-ts-eks/tests/index.ts
@@ -1,0 +1,48 @@
+// Copyright 2016-2020, Pulumi Corporation.  All rights reserved.
+
+import * as aws from "@pulumi/aws";
+import * as policy from "@pulumi/policy";
+import * as pulumi from "@pulumi/pulumi";
+
+const stackPolicy: policy.StackValidationPolicy = {
+    name: "eks-test",
+    description: "EKS integration tests.",
+    enforcementLevel: "mandatory",
+    validateStack: async (args, reportViolation) => {
+        const clusterResource = args.resources.find(r => r.isType(aws.eks.Cluster));
+        const cluster = clusterResource && clusterResource.asType(aws.eks.Cluster);
+        if (!cluster) {
+            reportViolation("EKS Cluster not found");
+            return;
+        }
+
+        if (cluster.version !== "1.13") {
+            reportViolation(`Expected EKS Cluster '${cluster.name}' version to be '1.13' but found '${cluster.version}'`);
+        }
+
+        const vpcId = cluster.vpcConfig.vpcId;
+        if (!vpcId) {
+            if (!pulumi.runtime.isDryRun()) {
+                reportViolation(`EKS Cluster '${cluster.name}' has unknown VPC`);
+            }
+            return;
+        }
+
+        const ec2 = new aws.sdk.EC2({region: aws.config.region});
+        const response = await ec2.describeVpcs().promise();
+        const defaultVpc = response.Vpcs?.find(vpc => vpc.IsDefault);
+        if (!defaultVpc) {
+            reportViolation("Default VPC not found");
+            return;
+        }
+
+        if (defaultVpc.VpcId === vpcId) {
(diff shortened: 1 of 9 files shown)


testing/pac-ts-eks/tests/index.ts
```diff
@@ -0,0 +1,48 @@
...
+            return;
+        }
+
+        if (cluster.version !== "1.13") {
+            reportViolation(`Expected EKS Cluster '${cluster.name}' version to be '1.13' but found '${cluster.version}'`);
+        }
+
+        const vpcId = cluster.vpcConfig.vpcId;
+        if (!vpcId) {
+            if (!pulumi.runtime.isDryRun()) {
```
leezen, line 25: Might be worth adding a comment here why we check this so that example readers understand what's going on here.

testing/pac-ts-eks/tests/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import * as aws from "@pulumi/aws";
+import * as policy from "@pulumi/policy";
+import * as pulumi from "@pulumi/pulumi";
+
+const stackPolicy: policy.StackValidationPolicy = {
+    name: "eks-test",
+    description: "EKS integration tests.",
+    enforcementLevel: "mandatory",
+    validateStack: async (args, reportViolation) => {
+        const clusterResource = args.resources.find(r => r.isType(aws.eks.Cluster));
```
leezen, line 12: Would it make sense to validate that there's only one cluster? Otherwise, other clusters would bypass all these checks.

testing/pac-ts-eks/tests/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+        const vpcId = cluster.vpcConfig.vpcId;
+        if (!vpcId) {
+            if (!pulumi.runtime.isDryRun()) {
+                reportViolation(`EKS Cluster '${cluster.name}' has unknown VPC`);
+            }
+            return;
+        }
+
+        const ec2 = new aws.sdk.EC2({region: aws.config.region});
```
leezen, line 31: I haven't tried this, but can't we use `.getProvider()` and then get the region from there? This way, we use the resource's region rather than rely on `aws.config.region` which may not necessarily be the same? It happens to be so in this case but that assumes some level of whitebox knowledge.

mikhailshilkov:
@leezen I'm sorry for a bit of a mess, but this code was already merged in another PR. I'll take your great comments and will apply in a new PR.

testing/pac-ts-eks/tests/index.ts
mikhailshilkov: So the `cluster` thing is not a proper resource but rather a property bag. An advantage of that is that there are no outputs to work around. But likewise, there's not `.getProvider()` to call directly. @justinvp do you know if there's an easy way to get the region of a given resource in PaC?

testing/pac-ts-eks/tests/index.ts
mikhailshilkov: Done in #623 and applies to the guide

testing/pac-ts-eks/tests/index.ts
mikhailshilkov: Done in #623 and applies to the guide

PR #587, opened 2020-03-03
Add 'aws-go-console-slack-notification' example.

clstokes:
Deploys a Lambda function and relevant CloudTrail and CloudWatch resources to send a 
Slack notification for any resource operation that is performed via the AWS Console.


aws-go-console-slack-notification/Makefile
```diff
@@ -0,0 +1,5 @@
+build::
+	GOOS=linux GOARCH=amd64 go build -o ./handler/dist/handler ./handler/handler.go
+	zip -j ./handler/dist/handler.zip ./handler/dist/handler
+dep::
+	dep ensure -v
```
EvanBoyle, line 5: Are you able to update this to go modules? I don't think we want to put out any more content referencing dep. It's considered a dead/stale tool by the community from what I understand.

clstokes: Yep, done.

PR #594, opened 2020-03-06
Add example for Go with Docker

tasiah:
For pulumi/pulumi-docker#128. Relies on pulumi/pulumi-docker#147.


azure-go-aci/main.go
```diff
@@ -0,0 +1,79 @@
...
+		portArgs := containerservice.GroupContainerPortArgs{
+			Port:     pulumi.Int(80),
+			Protocol: pulumi.String("TCP"),
+		}
+		containerArgs := containerservice.GroupContainerArgs{
+			Cpu:    pulumi.Float64(0.5),
+			Image:  image.ImageName,
+			Memory: pulumi.Float64(1.5),
+			Name:   pulumi.String("hello-world"),
+			Ports:  containerservice.GroupContainerPortArray([]containerservice.GroupContainerPortInput{portArgs}),
```
lukehoban: @pgavlin has pointed out to me that in general - this pattern can always be written as this instead:

```go
containerservice.GroupContainerPortArray({portArgs}),
```

Same applies in a few more places in this file (and in other Go examples we have).

tasiah: When I try `containerservice.GroupContainerPortArray({portArgs})`, Go says it's a syntax error: `syntax error: unexpected {, expecting expression`. Am I missing something?

lukehoban: Sorry - typed that wrong - it's:

```go
containerservice.GroupContainerPortArray{portArgs},
```

tasiah:
Should I wait to merge this until the next release of pulumi/pulumi-docker?

tasiah:
I realized I could just update the `Gopkg.toml` to use the latest revision (SHA) of `pulumi/pulumi-docker`. So after pulumi/pulumi-docker#149 gets merged, I'll update the `Gopkg.toml` to that SHA and merge this example (finally!).

PR #597, opened 2020-03-10
Convert node and python examples to Azure 2.0

mikhailshilkov:
Part of #598

### DIFF
--- a/azure-ts-appservice-devops/infra/index.ts
+++ b/azure-ts-appservice-devops/infra/index.ts
@@ -49,7 +49,7 @@ const blob = new azure.storage.Blob(`${prefix}-b`, {
     storageContainerName: storageContainer.name,
     type: "Block",
 
-    source: new pulumi.asset.FileArchive("../src"),
+    source: new pulumi.asset.FileArchive("../src/bin/Debug/netcoreapp2.1/publish"),
 });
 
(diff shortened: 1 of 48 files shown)


mikhailshilkov:
The only test failure is Spark. Azure support is still busy with it.

lukehoban:
> The only test failure is Spark. Azure support is still busy with it.

What do you think about disabling that test for now until we get a resolution from Azure support?

mikhailshilkov:
@lukehoban Yeah, that's probably the right thing to do. I'll open a PR.

azure-ts-appservice-devops/infra/index.ts
```diff
@@ -44,21 +44,20 @@ const storageContainer = new azure.storage.Container(`${prefix}-c`, {
...
-const blob = new azure.storage.ZipBlob(`${prefix}-b`, {
-    resourceGroupName: resourceGroup.name,
+const blob = new azure.storage.Blob(`${prefix}-b`, {
     storageAccountName: storageAccount.name,
     storageContainerName: storageContainer.name,
-    type: "block",
+    type: "Block",
 
-    content: new pulumi.asset.FileArchive("../src/bin/Debug/netcoreapp2.1/publish"),
+    source: new pulumi.asset.FileArchive("../src"),
```
lukehoban: Does this work?

PR #599, opened 2020-03-10
Skip HDInsights tests

mikhailshilkov:
An HDInsight cluster is currently stuck in our account, so we can't create new ones. Disable the tests temporarily.

Honestly, we had too much problems with HDInsights and never found any useful regression with those, so I wouldn't mind just removing these two tests permanently.


mikhailshilkov:
I'll merge as-is, hoping to convert these tests to "preview-only" as described in https://github.com/pulumi/pulumi/issues/3762

PR #601, opened 2020-03-11
feat: add go sample for API Gateway

retgits:
As discussed with @stack72, here's a PR that adds a Go sample for an AWS Lambda function with API Gateway. This sample extends the existing aws-go-lambda sample with new functionality:

- Adding an API Gateway
- Adding an API Gateway method and resource
- Adding a trigger to the AWS Lambda function
- Adding an API Gateway Deployment

The result is a deployed API Gateway, which sends messages to the AWS Lambda function and returns the result.

```bash
$ curl https://<gateway-id>.execute-api.us-west-2.amazonaws.com/prod/helloworld   
HELLOWORLD
```

### DIFF
--- a/aws-go-lambda-gateway/Gopkg.toml
+++ b/aws-go-lambda-gateway/Gopkg.toml
@@ -1,0 +1,42 @@
+# Gopkg.toml example
+#
+# Refer to https://golang.github.io/dep/docs/Gopkg.toml.html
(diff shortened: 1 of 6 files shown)


stack72:
Thanks for your work here @retgits :) I made 1 change locally and then pushed the changes! This is an awesome example!

PR #607, opened 2020-03-18
Update C# examples to stack-based API

### LINKED ISSUE #598: Convert all examples to Azure 2.0

### LINKED ISSUE #606: Convert all C# examples to stack-based API

mikhailshilkov:
- Update all .NET examples to the latest versions of NuGet
- Include all of them to the solution file
- Refactor to stack-based APIs

Resolves #598. Resolves #606.


aws-cs-lambda/README.md
```diff
@@ -28,7 +28,7 @@ pulumi up -C ./pulumi
 3. Call our lambda function from the aws cli.
 ```bash
 aws lambda invoke \
---function-name $(pulumi stack output lambda -C ./pulumi) \
+--function-name $(pulumi stack output Lambda -C ./pulumi) \
```
lukehoban: I don’t think I had realized this implication of the stack approach - the stack outputs now will be PascalCase by default. That’s leads to a slightly surprising inconsistency with other languages. Out of curiosity, can you pass an alternative name to the output attribute to have this be seen externally by a different name that’s the .NET property?

mikhailshilkov: Yes, one can do `[Output("myName")]`. I actually started with lowercasing by default but Cyrus made a good point that if would be surprising for .NET folks.

lukehoban: Lowercasing by default I can see being surprising.

But I do wonder if as a style thing we should have all our examples map these to lowercase names explicitly (with `[Output("myName")]`)?

mikhailshilkov: What would be the upside of that? I don't expect users to want to do so that often?

lukehoban: Ultimately just some amount of consistency in export naming across different languages (the names of stack exports really shouldn't be dependent on what language it was implemented in).

But I'm fine with this as is - just another little language inconsistency.

mikhailshilkov: I think I'll merge as-is and will think about a good place to show the capability of explicit names

PR #617, opened 2020-03-30
Move AWS examples to go modules

EvanBoyle:


aws-go-fargate/README.md
```diff
@@ -31,16 +31,10 @@ Next, to deploy the application and its infrastructure, follow these steps:
...
 
     ```bash
-    $ dep ensure
-    ```
-
-4. Compile your Go application:
-
-    ```bash
-    $ go build .
+    $ go mod download
```
lukehoban: Users don't actually have to do this do they?  It gets automatically done as needed.  I think we can leave out of instructions.

EvanBoyle: Good point. go run and go build both acquire dependencies automatically. Removed.

EvanBoyle:
Test failure is unrelated to my change. Also happening in the current master:

https://travis-ci.com/github/pulumi/examples/jobs/308021387

Going to merge.

PR #618, opened 2020-03-31
Update gcp & azure examples to go modules

EvanBoyle:
Some of them had code errors that I fixed. I ran a `pulumi preview` on all of them after the update to verify.


EvanBoyle:
fixes https://github.com/pulumi/pulumi/issues/3817

azure-go-webserver-component/main.go
```diff
@@ -41,6 +41,14 @@ func main() {
 			},
 		})
 
+		subnetID := network.Subnets.Index(pulumi.Int(0)).Id().ApplyT(func(val interface{}) (string, error) {
+			v := val.(*string)
```
lukehoban: Total nit, but can this not be `func(v *string)`?

EvanBoyle: Good catch. Updated.

EvanBoyle:
Unrelated build failure. `TestAccPacketPyWebserver` has been failing in the master cron build for some time.

PR #619, opened 2020-03-31
Create an example for an Azure Function (as container) on a Linunx AppServicePlan

StefH:


mikhailshilkov:
Hi @StefH 
Thanks for your PR!
I'd love to merge your example with https://github.com/pulumi/examples/tree/master/azure-cs-functions-raw so that it would show both options in one place similar to what we do in https://github.com/pulumi/examples/tree/master/azure-ts-functions-raw.

A possible plan:
- Bring the `ArchiveFunctionApp` component from [here](https://github.com/pulumi/examples/blob/master/azure-cs-cosmosapp-component/ArchiveFunctionApp.cs)
- Adjust it if needed
- Use it two times to deploy two function apps on Windows and Linux

What do you think?

StefH:
- I'm not sure I fully understand that `ArchiveFunctionApp`, this is more an advanced scenario which I  will not use for this example
- What is the original example named `***-raw` ? What means `raw`? My intention is just to remove the word `raw` from folder + project.
- How do you  mean deploy it two times? I guess this has to with converting the function into a component, however that seems complicated. My idea of joining was just to create 2 different example service plans and deploy two functions

mikhailshilkov:
Yes, `ArchiveFunctionApp` is the component that deploys all resources required to deploy a Function App: storage, container, blob, plan, and app. It has parameters for runtime and may have a parameter for OS.

The word `raw` is a bit unfortunate, we may remove it. It came from TypeScript where it means to oppose such functions to callback-based functions embedded into a Pulumi program.

There are many useful permutations of a Function App configuration, e.g. the TS counterpart example currently has 6 apps deployed. That's why I'm trying to avoid spreading similar things into multiple distinct examples.

Why do you find a component complicated? Do you see a way to make it clearer?

StefH:
My personal opinion is that examples should be easy to understand, and that there should be multiple complexity levels. Not combining all complexity into 1 example.

Building a Linux App Service based function, tool me a lot of work and retries before it worked as expected (https://github.com/StefH/AzureFunctionsV3/blob/master/Pulumi/AzureFunctionStack.cs)

There is no harm in duplicating example code if the purpose differs.

So actually, there should be more examples listed (without the 'raw' postfix)
- Function on Windows Service Plan
- Function on Linux Service Plan
- Function on Consumption Plan

And all examples should have the correct (= only required) app-settings and have application-insights configured.

Probably only the latest version V3 should be defined, however I don't know if V1 and V2 can also be supported by Pulumi.

mikhailshilkov:
@StefH sorry for a long pause: I've been busy with 2.0 bits.

I couldn't change your PR, so I opened a new one. What I did:

- Update to the latest version
- Use Python instead of .NET to show off different runtimes in different examples
- Renamed both examples
- Added to readme

Do you mind taking a look at #664?

StefH:
@mikhailshilkov I will take a look at #664

mikhailshilkov:
Thanks for a useful contribution! This has been merged in #664

PR #621, opened 2020-04-01
Update the Go unit test example

mikhailshilkov:
Split out of #615 for faster turnaround

### DIFF
--- a/testing-unit-go/main.go
+++ b/testing-unit-go/main.go
@@ -1,25 +1,69 @@
 package main
 
 import (
-	"github.com/pulumi/pulumi-aws/sdk/go/aws/s3"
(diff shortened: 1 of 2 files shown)


mikhailshilkov:
> Does this only work in us-east-2 (due to the AMI)? If so should we update the readme?

Well, this is a unit test with mocks, so the test doesn't care about the region as it does not do anything "real" with AWS.

EvanBoyle:
The example is still deployable though no? I guess the readme doesn't specify any instructions on deploying the app.

mikhailshilkov:
It hopefully is, but I focused README's of all unit tests on simply running the tests. If you think we should improve that, do you mind opening a separate issue? I don't think I can get to this today. Thanks!

mikhailshilkov:
Merging as this example has no effect on the build

PR #624, opened 2020-04-04
Add Kubernetes Go example to the index

joeduffy:
This will allow us to link to it from our nurturing (which
currently says "Coming Soon!")

Also fixed up some formatting issues, like capitalizing "script"
in TypeScript and JavaScript, and getting rid of some tabs.


joeduffy:
FYI @zchase , will update the nurturing flow to link to the new Kubernetes Go anchor.

PR #628, opened 2020-04-06
Update the packet examples to remove ipAddressTypes

stack72:


packet-py-webserver/__main__.py
```diff
@@ -16,7 +16,6 @@
     operating_system="coreos_stable",
     plan="baremetal_0",
     project_id=project.id,
-    ip_address_types=["public_ipv4", "private_ipv4"],
```
lukehoban, line 19: Does this work?  I suspect they will no longer have public IP addresses after this (making the example not work as intended)?

stack72: Made it more explicit

PR #630, opened 2020-04-06
Add examples for secrets providers

jaxxstorm:
This introduces a simple example for setting up secrets providers for each of the supported clouds. I've tried to implement this in a way that makes sense for the repo, but happy to change the directory layout if needed.

### DIFF
--- a/secrets-provider/azure/Pulumi.yaml
+++ b/secrets-provider/azure/Pulumi.yaml
@@ -1,0 +1,3 @@
+name: pulumi-azure-kms
+runtime: nodejs
+description: Minimal config which shows how Azure encryption support works
--- a/secrets-provider/azure/Pulumi.yaml
+++ b/secrets-provider/azure/Pulumi.yaml
@@ -1,3 +1,3 @@
-name: pulumi-azure-kms
+name: pulumi-azure-keyvault
 runtime: nodejs
 description: Minimal config which shows how Azure encryption support works
--- a/secrets-provider/azure/README.md
+++ b/secrets-provider/azure/README.md
@@ -1,12 +1,12 @@
 # Pulumi Azure KMS Encryption
 
-Pulumi allows you to use KMS encryption from your cloud provider to encrypt any secrets stored in the backend.
+Pulumi allows you to use Azure Keyvault encryption from your cloud provider to encrypt any secrets stored in the backend.
 
 This example shows how this might be done for Azure Keyvault. It creates a storage bucket with a single file that has a "secret" value.
 
 # Getting Started
 
-To use this example, perform the following steps. This examples assumes you have the pulumi-cli installed and the gcloud SDK installed.
+To use this example, perform the following steps. This examples assumes you have the pulumi-cli installed and the Azure CLI installed.
 
 You should also ensure:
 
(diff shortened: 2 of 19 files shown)


stack72:
Oh I'm a big fan of this!

secrets-provider/azure/README.md
```diff
@@ -0,0 +1,118 @@
+# Pulumi Azure KMS Encryption
+
+Pulumi allows you to use KMS encryption from your cloud provider to encrypt any secrets stored in the backend.
+
+This example shows how this might be done for Azure Keyvault. It creates a storage bucket with a single file that has a "secret" value.
+
+# Getting Started
+
+To use this example, perform the following steps. This examples assumes you have the pulumi-cli installed and the gcloud SDK installed.
```
stack72, line 9: Not gcloud SDK

secrets-provider/azure/README.md
```diff
@@ -0,0 +1,118 @@
+# Pulumi Azure KMS Encryption
+
+Pulumi allows you to use KMS encryption from your cloud provider to encrypt any secrets stored in the backend.
```
stack72, line 3: It's not KMS - it's KeyVault

secrets-provider/azure/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: pulumi-azure-kms
```
stack72, line 1: should be keyvault

secrets-provider/azure/README.md
```diff
@@ -0,0 +1,118 @@
...
+# Getting Started
+
+To use this example, perform the following steps. This examples assumes you have the pulumi-cli installed and the gcloud SDK installed.
+
+You should also ensure:
+
+  * You azure command line tool installed
+  * You are logging in via the `az` command line tool.
+  * You have created a resource-group
+  * You must have the environment variable `AZURE_KEYVAULT_AUTH_VIA_CLI` set to `true` eg: `export AZURE_KEYVAULT_AUTH_VIA_CLI=true`
```
stack72, line 16: I think we should also suggest that we can do this via environment variables as we can set some for GoCloud AFAICR

secrets-provider/azure/README.md
```diff
@@ -0,0 +1,118 @@
...
+pulumi config set bucketName pulumi-lbriggs
+pulumi config set --secret secretValue "correct-horse-battery-stable"
+```
+
+## Create the stack
+
+```bash
+# This will create the stack without prompting, be aware!
+pulumi up --yes                                                                                                  home.lbrlabs/default ⎈
+Previewing update (gcloud-kms):
```
stack72, line 67: I think you are mixing up the outputs - this seems to be of gcloud

secrets-provider/azure/README.md
jaxxstorm: Fixed! Copy paste fail, thanks for spotting!

secrets-provider/azure/Pulumi.yaml
jaxxstorm: Fixed!

secrets-provider/azure/README.md
jaxxstorm: Looks like a copy paste fail again, have fixed this! Thanks!

PR #631, opened 2020-04-06
Adding an example that deploys an application to an EKS cluster

stack72:


stack72:
Just added the README - thanks for verifying this :)

PR #646, opened 2020-04-14
Bump dependencies for all Go examples for v2

jaxxstorm:
This only deals with Go dependencies for now.

Draft because we need to wait for these to be published


azure-go-aci/go.mod
```diff
@@ -4,6 +4,6 @@ go 1.13
 
 require (
 	github.com/pulumi/pulumi-azure/sdk/v2 v2.3.1
```
EvanBoyle, line 6: This and the other azure SDKs need to be bumped to 3.0

gcp-go-functions-raw/go.mod
```diff
@@ -4,5 +4,5 @@ go 1.13
 
 require (
 	github.com/pulumi/pulumi-gcp/sdk/v2 v2.11.1
```
EvanBoyle, line 6: GCP SDKs also need to be bumped from 2.0 -> 3.0

PR #648, opened 2020-04-15
Update .NET examples to 2.0 betas

mikhailshilkov:


mikhailshilkov:
The test failure is DigitalOcean/python - unrelated

PR #650, opened 2020-04-15
Migrate from Travis to GitHub Actions

stack72:
This migrates the PR and master pushes of the examples workflow to use GitHub actions

the cron has not been added right now so therefore I am leaving that *temporarily* on Travis

This PR will continue to evolve to not be so install heavy and more than likely get pushed into a build container - this is just the initial work to allow the migration


.github/workflows/ci.yml
```diff
@@ -0,0 +1,123 @@
...
+          python-version: ${{ matrix.python-version }}
+      - name: Install Python Deps
+        run: |
+          pip3 install pipenv
+      - name: Install aws-iam-authenticator
+        run: |
+          curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.13.7/2019-06-11/bin/linux/amd64/aws-iam-authenticator
+          chmod +x ./aws-iam-authenticator
+          sudo mv aws-iam-authenticator /usr/local/bin
+      - name: Install Kubectl
```
jaxxstorm, line 70: Is it worth switching to this action for this?
https://github.com/Azure/setup-kubectl ?

jaxxstorm:
Love this!

.github/workflows/ci.yml
stack72: There's a tonne of refactoring I want to do here - so we can add this to our list to refactor

PR #675, opened 2020-04-30
Add a Python provisioning example

joeduffy:
This adds an example of using SSH to configure an EC VM in Python (i.e., "provisioning"). It's effectively the Python equivalent to https://github.com/pulumi/examples/tree/master/aws-ts-ec2-provisioners and includes example dynamic resources that copy files and execute commands.


joeduffy:
@lukehoban @stack72 Aiming to get back to a customer with this example tonight -- please let me know if you have feedback and I'll incorporate it retroactively.

Issue #634, opened 2020-04-07
Guestbook examples are not working


lukehoban:
Most of our Guestbook examples are not working correctly today, as the image we run expects a service on the network called `redis-slave`, but we run a service with the name `redis-replica` (or in some cases `redis-replica-01ab23cd`).

Although the Guestbook page loads, submitting an entry via the HTML input box fails to do anything, and reports an error in the JavaScript console.

We should ensure that all Guestbook examples do something similar to the C# example:

```cs
        var redisReplicaService = new Pulumi.Kubernetes.Core.V1.Service("redis-replica", new ServiceArgs
        {
            Metadata = new ObjectMetaArgs
            {
                Name = "redis-slave",
                Labels = redisReplicaDeployment.Metadata.Apply(metadata => metadata.Labels)
            },
```

We should fix this across `simple` and `components` examples for NodeJS, Python and Go as well.

stack72:
ok, these have been tested individually (using pulumi up and posting messages to the UI). I also have added these to the PR tests for the examples repo

```
--- PASS: TestAccKubernetesGuestbook (1439.93s)
    --- PASS: TestAccKubernetesGuestbook//Users/stack72/code/go/src/github.com/pulumi/examples/kubernetes-go-guestbook/simple (217.25s)
        examples_test.go:1377:
    --- PASS: TestAccKubernetesGuestbook//Users/stack72/code/go/src/github.com/pulumi/examples/kubernetes-go-guestbook/components (217.46s)
        examples_test.go:1377:
    --- PASS: TestAccKubernetesGuestbook//Users/stack72/code/go/src/github.com/pulumi/examples/kubernetes-cs-guestbook/simple (199.31s)
        examples_test.go:1377:
    --- PASS: TestAccKubernetesGuestbook//Users/stack72/code/go/src/github.com/pulumi/examples/kubernetes-cs-guestbook/components (215.04s)
        examples_test.go:1377:
    --- PASS: TestAccKubernetesGuestbook//Users/stack72/code/go/src/github.com/pulumi/examples/kubernetes-py-guestbook/simple (186.15s)
        examples_test.go:1377:
    --- PASS: TestAccKubernetesGuestbook//Users/stack72/code/go/src/github.com/pulumi/examples/kubernetes-ts-guestbook/simple (200.43s)
        examples_test.go:1377:
[...]

PR #678, opened 2020-05-04
Fixing up Kubernetes Guestbook Examples

stack72:
Fixes: #634

Also adds the ability to run the tests as part of our CI tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,3 +123,205 @@ jobs:
         run: make install
       - name: Lint typescript files
         run: make lint
+  test-infra-setup:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: [ubuntu-latest]
+        go-version: [1.13.x]
+        node-version: [10.x]
+        python-version: [3.7]
+        dotnet-version: ['3.1.100']
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - name: Install DotNet ${{ matrix.dotnet-version }}
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: ${{ matrix.dotnet-version }}
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Set up Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v1
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Python Deps
+        run: |
+          pip3 install pipenv
+      - name: Install aws-iam-authenticator
+        run: |
+          curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/1.13.7/2019-06-11/bin/linux/amd64/aws-iam-authenticator
+          chmod +x ./aws-iam-authenticator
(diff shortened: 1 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -95,30 +185,142 @@ jobs:
...
-      - name: Run ${{ matrix.tests-set }} Tests
-        run: make specific_test_set TestSet=${{ matrix.tests-set }}
-  lint:
+      - name: Create test infra
+        run: |
+          export STACK=${{ secrets.PULUMI_TEST_OWNER }}/ci-cluster-$(head /dev/urandom | LC_CTYPE=C tr -dc '[:lower:]' | head -c5)
+          echo "::set-env name=STACK::$STACK"
+          make setup_test_infra
+
+  test-infra-destroy:
```
lukehoban, line 194: I'm a little confused overall in what's going on in this file now. Is this step really about destroying?  Why do we have all the steps copied twice?  Why does `kubernetes` need a separate block?

Can we keep all this simpler?

misc/scripts/destroy-ci-cluster.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -o nounset -o errexit -o pipefail
+
+echo Deleting ephemeral Kubernetes cluster...
+
+pushd misc/scripts/testinfra/
+pulumi stack select "${STACK}" && \
```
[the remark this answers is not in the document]
stack72: this is a direct copy of the work to allow a test setup in the pulumi-kubernetes repo so we can definitely change this for the better :)

.github/workflows/ci.yml
```diff
@@ -32,11 +32,101 @@ env:
...
+#      - name: Install Yarn
+#        run: curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version ${{ matrix.yarn-version }}
+#      - name: Update PATH for Yarn
+#        run: |
+#          echo "::add-path::$HOME/.yarn/bin"
+#          echo "::add-path::$HOME/.config/yarn/global/node_modules/.bin"
+#      - name: setup linting tool
+#        run: make install
+#      - name: Lint typescript files
+#        run: make lint
```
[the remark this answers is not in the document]
stack72: yes - sorry, this is a WIP and I just marked it as so

.github/workflows/ci.yml
stack72: This *might* be able to get simpler - I might even break out Kubernetes to a different workflow file. The reason it's a separate block is that we want to be able to run AWS, GCP, Azure etc and at the same time, spin up our test infra for the Kubernetes tests

We can definitely do something to make this more obvious of course

stack72:
I am going to suggest that this is a good first step! This will be refactored as part of our build and deploy changes - but right now, this at least gets some of the Kubernetes examples on the map

@lukehoban are you willing to let this be a good intermediate? :)

kubernetes-cs-guestbook/simple/Guestbook.cs
```diff
@@ -256,7 +256,7 @@ public Guestbook()
         }
         else
         {
-            this.FrontendIp = frontendService.Status.Apply(status => status.LoadBalancer.Ingress[0].Hostname);
+            this.FrontendIp =frontendService.Status.Apply(status => status.LoadBalancer.Ingress[0].Ip ?? status.LoadBalancer.Ingress[0].Hostname);
```
lukehoban, line 259:
```suggestion
            this.FrontendIp = frontendService.Status.Apply(status => status.LoadBalancer.Ingress[0].Ip ?? status.LoadBalancer.Ingress[0].Hostname);
```

PR #691, opened 2020-05-18
.NET unit testing examples for the blog post

mikhailshilkov:


testing-unit-cs-mocks/Testing.cs
```diff
@@ -0,0 +1,69 @@
...
+            var outputs = ImmutableDictionary.CreateBuilder<string, object>();
+            
+            // Forward all input parameters as resource outputs, so that we could test them.
+            outputs.AddRange(inputs);
+            
+            // Set the name to resource name if it's not set explicitly in inputs.
+            if (!inputs.ContainsKey("name"))
+                outputs.Add("name", name);
+            
+            if (type == "azure:storage/blob:Blob")
```
stack72, line 24: would this not be better served with a switch? It seems long winded to continually add if statements

testing-unit-fs-mocks/Testing.fs
```diff
@@ -0,0 +1,35 @@
...
+open System.Threading.Tasks
+open Pulumi.Testing
+
+type Mocks() =
+    interface IMocks with
+        member this.NewResourceAsync(typeName: string, name: string, inputs: ImmutableDictionary<string, obj>, provider: string, id: string): Task<ValueTuple<string,obj>> =
+            let inputKVs = inputs |> Seq.map(fun kv -> kv.Key, kv.Value) |> Seq.toList
+            let nameKVs = if inputs.ContainsKey("name") then [] else [("name", name :> obj)]
+            let endpointKVs =
+                if typeName = "azure:storage/account:Account"
```
stack72, line 17: Do we not need the same removal for source from the Blob as per the CSharp example?

mikhailshilkov: It happens (in slightly weird way to avoid mutations) on line 24

testing-unit-cs-mocks/Testing.cs
mikhailshilkov: You are right, but it's are easier to gradually add/explain `if`'s in the blog post. I'll change to switch next time I revisit this example after the blog is published.

stack72: 👍

testing-unit-fs-mocks/Testing.fs
stack72: ah... ok :) That's an Fsharp-ism that I didn't get :)

PR #693, opened 2020-06-04
Add S3 folder examples for C# and F#

mikhailshilkov:
I use these examples for quick demos, I figured I should publish them


mikhailshilkov:
The test failures seem unrelated, at least AWS C#/F# ones are green

Issue #684, opened 2020-05-11
Repo examples status from GHA back to Slack


stack72:
We need to report on a failed build

PR #696, opened 2020-06-08
Ensuring we notify slack on build failures

stack72:
Fixes: #684


stack72:
This is also going to change to be under codegen so we can swap it out when GH allows meta actions ;)

Issue #695, opened 2020-06-05
update helm examples to use fetchOpts


jaxxstorm:
Many of our examples tell the user to add the helm repo using `helm repo add` before running a pulumi program. We should switch these to `fetchOpts` with the repo. Link:

https://github.com/search?q=org%3Apulumi+%22helm+repo+add%22&type=Code

PR #697, opened 2020-06-08
Update Helm Chart examples to use fetchOpts to pass the repository

stack72:
Fixes: #695


stack72:
Failures are unrelated and are being worked on separately

PR #704, opened 2020-06-11
Add azure provisioner example

EvanBoyle:
Ports our ec2 provisioners example over to Azure. 

I had to use an extremely permissive security group. The ssh connection was timing out when I tried to use targeted rules on port 22.


azure-ts-vm-provisioners/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: aws-ts-ec2-provisioners
```
ericrudder, line 1: maybe give this example one more set of review to make sure it's Azure and not AWS  (several places ... )

azure-ts-vm-provisioners/provisioners/index.ts
```diff
@@ -0,0 +1,204 @@
+// Copyright 2016-2019, Pulumi Corporation.  All rights reserved.
```
ericrudder, line 1: all (c) should be 2020

PR #734, opened 2020-06-29
converted aws-go-appsync from python to golang

sashu-shankar:
Conversion of aws-py-appsync to Golang => aws-go-appsync


misc/test/examples_test.go
```diff
@@ -223,6 +223,15 @@ func TestAccAwsPyAppSync(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestAccAwsGoAppSync(t *testing.T) {
+	test := getAWSBase(t).
+		With(integration.ProgramTestOptions{
+			Dir: path.Join(getCwd(t), "..", "..", "aws-go-appsync"),
+		})
```
EvanBoyle, line 230: you can also include additional runtime validation here to make this test more robust. You can translate that curl command into an HTTP request, some of the other examples already do this in their tests: https://github.com/pulumi/examples/blob/1c721ea2435478165850f041d14bdfd50c29d2d4/misc/test/examples_test.go#L48

aws-go-appsync/main.go
```diff
@@ -0,0 +1,195 @@
...
+
+		// Link the data source to the Dynamo DB Table
+		dataSource, err := appsync.NewDataSource(ctx, "tenants-DS", &appsync.DataSourceArgs{
+			Name:  randStr.Result,
+			ApiId: api.ID(),
+			Type:  pulumi.String("AMAZON_DYNAMODB"),
+			DynamodbConfig: &appsync.DataSourceDynamodbConfigArgs{
+				TableName: table.Name,
+			},
+			ServiceRoleArn: role.Arn.ApplyString(func(arn string) string {
```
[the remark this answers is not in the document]
sashu-shankar: That worked!

misc/test/examples_test.go
```diff
@@ -227,8 +227,25 @@ func TestAccAwsGoAppSync(t *testing.T) {
 	test := getAWSBase(t).
 		With(integration.ProgramTestOptions{
 			Dir: path.Join(getCwd(t), "..", "..", "aws-go-appsync"),
-		})
+			ExtraRuntimeValidation: func(t *testing.T, stack integration.RuntimeValidationStackInfo) {
+				maxWait := 15 * time.Minute
 
+				endpoint := stack.Outputs["endpoint"].(string)
+				queryParams := "?query=mutation%20AddTenant%20%7B%20addTenant%28id%3A%20%22123%22%2C%20name%3A%20%22FirstCorp%22%29%20%7B%20id%20name%20%7D%20%7D"
```
EvanBoyle, line 234: can we use the plain query here in combination with `net/url.QueryEscape` to make things easier to read? 

https://golang.org/pkg/net/url/#QueryEscape

misc/test/examples_test.go
```diff
@@ -227,8 +227,25 @@ func TestAccAwsGoAppSync(t *testing.T) {
 	test := getAWSBase(t).
 		With(integration.ProgramTestOptions{
 			Dir: path.Join(getCwd(t), "..", "..", "aws-go-appsync"),
-		})
+			ExtraRuntimeValidation: func(t *testing.T, stack integration.RuntimeValidationStackInfo) {
+				maxWait := 15 * time.Minute
```
EvanBoyle, line 231: 15 minutes seems like a long time. How long does it typically take for this test to pass? I'd guess 1-2 minutes. Taking that average value and doubling or tripling it should be fine. We can always extend the timeout later. 

Long timeouts should generally be avoided if possible as they extend the overall time it takes for a test run to fail if there actually is an error in the test. (Consider if there is a regression and the ready function continuously fails).

misc/test/examples_test.go
```diff
@@ -227,8 +227,25 @@ func TestAccAwsGoAppSync(t *testing.T) {
...
+				endpoint := stack.Outputs["endpoint"].(string)
+				queryParams := "?query=mutation%20AddTenant%20%7B%20addTenant%28id%3A%20%22123%22%2C%20name%3A%20%22FirstCorp%22%29%20%7B%20id%20name%20%7D%20%7D"
+
+				key := stack.Outputs["key"].(string)
+				headersMap := map[string]string{
+					"Content-Type": "application/graphql",
+					"x-api-key":    key,
+				}
+
+				assertHTTPResultShapeWithRetry(t, endpoint+queryParams, headersMap, maxWait, func(body string) bool {
```
EvanBoyle, line 242: The idiomatic way to concat strings in go is using the `fmt` package (Sprintf in this case). 

Something like:

```go
endpoint := stack.Outputs["endpoint"].(string)
mutation := "mutation AddTenant { addTenant(id: \"123\", name: \"FirstCorp\") { id name } }"
getURL := fmt.Sprintf("%s?query=%s", endpoint, url.QueryEscape(mutation))

assertHTTPResultShapeWithRetry(t, getUrl, ...
```

misc/test/examples_test.go
sashu-shankar: Just shortened it to 8 min

sashu-shankar:
The test for Aws, Go is currently inactive as the build was taking a large amount of memory as it was running 3-4 tests in parallel, thus are not passing. To verify the changes made, I ran the test locally to ensure it is passing.

PR #738, opened 2020-07-02
converted aws assume-role to go, add README.md, and the test for it

zephyrz73:
1. Fixing the line break in README.md of aws-py-serverless-raw
2. Converted aws-ts-assume-role to aws-go-assume-role
3. Add test to misc/test/examples_test.go


zephyrz73:
Seems like the dotnet publish command functionality is still under development. The test for aws-ts/py-serverless-raw is skipped in that case. The assume-role test works fine.

aws-go-assume-role/assume-role/main.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+func main() {
+	pulumi.Run(func(ctx *pulumi.Context) error {
+
+		config := config.New(ctx, "")
+		roleToAssumeARN := config.Require("roleToAssumeARN")
+
+		region := GetRegion(ctx)
```
justinvp, line 25: I haven't tried this, but I think it would allow your `GetRegion` function to be deleted:


```suggestion
		region := config.Require(ctx, "aws:region")

		c := config.New(ctx, "")
		roleToAssumeARN := c.Require("roleToAssumeARN")
```

misc/test/examples_test.go
```diff
@@ -247,6 +260,14 @@ func TestAccAwsPyS3Folder(t *testing.T) {
...
 
 func TestAccAwsPyServerlessRaw(t *testing.T) {
+	cmd := exec.Command("dotnet publish", "app")
+	cmd.Path = path.Join(getCwd(t), "..", "..", "aws-ts-serverless-raw", "app")
+	cmd.Dir = path.Join(getCwd(t), "..", "..", "aws-ts-serverless-raw")
+	err := cmd.Run()
+	if err != nil {
+		fmt.Printf("err: %v", err)
+		return
+	}
```
justinvp, line 270: From conversation in Slack I think this needs to be deleted until https://github.com/pulumi/pulumi/pull/4559 is available and can be used instead.

misc/test/examples_test.go
```diff
@@ -491,6 +512,14 @@ func TestAccAwsTsS3LambdaCopyZip(t *testing.T) {
...
 
 func TestAccAwsTsServerlessRaw(t *testing.T) {
+	cmd := exec.Command("dotnet", "publish")
+	cmd.Path = path.Join(getCwd(t), "..", "..", "aws-ts-serverless-raw", "app")
+	cmd.Dir = path.Join(getCwd(t), "..", "..", "aws-ts-serverless-raw")
+	err := cmd.Run()
+	if err != nil {
+		fmt.Printf("err: %v", err)
+		return
+	}
```
justinvp, line 522: From conversation in Slack I think this needs to be deleted until https://github.com/pulumi/pulumi/pull/4559 is available and can be used instead.

aws-go-assume-role/assume-role/main.go
zephyrz73: Changed it with new pull request with new conversions

misc/test/examples_test.go
zephyrz73: deleted

justinvp:
@zephyrz73, is https://github.com/pulumi/examples/pull/744 meant to replace this PR? If so, should this be closed?

PR #740, opened 2020-07-04
added hyperlink to repository

Zaubeerer:
I am a Pythonista new to pulumi and was at first confused that the repository was "forgotten to be added as hyperlink".

However, I then saw that on the top of the article there is a link to "View Code".

Probably it's easiest to just write "clone the code from the above mentioned code ("View Code").

Alternatively, adding a hyperlink as mentioned in this PR might be the most user-friendly solution.

Thanks for the great tool!

Bests
Robin

### DIFF
--- a/aws-py-fargate/README.md
+++ b/aws-py-fargate/README.md
@@ -14,7 +14,7 @@ This example is inspired by [Docker's Getting Started Tutorial](https://docs.doc
 
 ## Running the Example
 
-Clone this repo and `cd` into it.
+Clone [the examples repo](https://github.com/pulumi/examples/tree/master/aws-py-fargate) and `cd` into it.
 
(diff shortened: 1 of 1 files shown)


komalali:
Thanks for bringing this up @Zaubeerer 

For reference, this is the link to where the README code shows up on the docs site: https://www.pulumi.com/docs/tutorials/aws/aws-py-fargate/#running-the-example

Looks like there's also some funky rendering of the codeblock in step 4.

@stack72 we should make this fix?

stack72:
@Zaubeerer / @komalali apologies for the slow reply here - this actually makes a lot of sense and is something that someone else has suggested on one of our other guides

Merging now

Thanks @Zaubeerer

PR #746, opened 2020-07-19
Use requireSecret instead of require for password

benmccann:


benmccann:
Done. Thanks for the review

PR #759, opened 2020-08-12
Created aws-py-voting-app example

jetvova:


aws-py-voting-app/README.md
```diff
@@ -0,0 +1,105 @@
...
+## Prerequisites
+
+1. [Install Pulumi](https://www.pulumi.com/docs/get-started/install/)
+1. [Configure Pulumi for AWS](https://www.pulumi.com/docs/intro/cloud-providers/aws/setup/)
+1. [Configure Pulumi for Python](https://www.pulumi.com/docs/intro/languages/python/)
+1. [Install Docker](https://docs.docker.com/engine/installation/)
+
+## Deploying and running the program
+
+1. Install dependencies (a `virtualenv` is recommended - see [Pulumi Python docs](https://www.pulumi.com/docs/intro/languages/python/)):
```
stack72, line 22: This should only be run after we set a virtualenv

aws-py-voting-app/__main__.py
```diff
@@ -0,0 +1,298 @@
...
+config = pulumi.Config()
+redis_password = config.require("redis-password")
+redis_port = 6379
+
+# The ECS cluster in which our application and databse will run
+app_cluster = aws.ecs.Cluster("app-cluster")
+
+# Creating a VPC and a public subnet
+app_vpc = aws.ec2.Vpc("app-vpc",
+    cidr_block="172.31.0.0/16",
```
stack72, line 19: Do you want to hardcode this? Or make it configurable and then set a default if not set?

aws-py-voting-app/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pulumi>=2.0.0,<3.0.0
+pulumi-aws>=2.0.0,<3.0.0
+pulumi-docker>=1.0.0,<3.0.0
```
stack72, line 3: we should make this `pulumi-docker>=2.0.0,<3.0.0`

aws-py-voting-app/README.md
jetvova: Done!

aws-py-voting-app/__main__.py
jetvova: I think that unlike other configurations, the CIDR block might not be as interesting for users to change. Since this is a private network, 172.31.0.0 should always work for every project without creating deployment conflicts (even if you deploy multiple stacks of this), and if users really want to change it, I think it's obvious for them where to do it.

aws-py-voting-app/requirements.txt
jetvova: Done!

PR #762, opened 2020-08-15
Update GCP GKE example

runlevel5:
## What

Replace default node pool with a new separately managed node pool resource

## Why

Using a non-default node pool is highly recommended practice.

### DIFF
--- a/gcp-ts-gke-hello-world/index.ts
+++ b/gcp-ts-gke-hello-world/index.ts
@@ -12,18 +12,35 @@ export const masterVersion = config.get("masterVersion") ||
 
 // Create a GKE cluster
 const cluster = new gcp.container.Cluster(name, {
-    initialNodeCount: 2,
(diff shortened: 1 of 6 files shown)


stack72:
Thanks so much for the work here - you are correct - this is a way better thing to do - would you be interested in updating the PR to do the same thing on the other GKE clusters? 🥇

runlevel5:
@stack72 I've amended the PR

stack72:
Thanks so much @runlevel5 

Giving this a shot right now

stack72:
Hi @runlevel5 

I just tested this locally and all works - I am going to make 1 change before merging though

I see, from the tests, that the kubernetes namespace and service etc get deployed before the NodePool has finished being created so I am going to add a dependsOn to the kubernetes resources to ensure the NodePool is created first

Paul

Issue #773, opened 2020-08-24
gcp-ts-k8s-ruby-on-rails-postgresql doesn't compile


lukehoban:
Attempting to deploy this example fails with:

```
    error: Running program '/Users/lukehoban/dd/pulumidemos/gcp-rails' failed with an unhandled exception:
    TSError: ⨯ Unable to compile TypeScript:
    config.ts(21,95): error TS2339: Property 'latestMasterVersion' does not exist on type 'Promise<GetEngineVersionsResult>'.

        at createTSError (/Users/lukehoban/dd/pulumidemos/gcp-rails/node_modules/ts-node/src/index.ts:261:12)
        at getOutput (/Users/lukehoban/dd/pulumidemos/gcp-rails/node_modules/ts-node/src/index.ts:367:40)
        at Object.compile (/Users/lukehoban/dd/pulumidemos/gcp-rails/node_modules/ts-node/src/index.ts:558:11)
        at Module.m._compile (/Users/lukehoban/dd/pulumidemos/gcp-rails/node_modules/ts-node/src/index.ts:439:43)
        at Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
        at Object.require.extensions.<computed> [as .ts] (/Users/lukehoban/dd/pulumidemos/gcp-rails/node_modules/ts-node/src/index.ts:442:12)
        at Module.load (internal/modules/cjs/loader.js:985:32)
        at Function.Module._load (internal/modules/cjs/loader.js:878:14)
        at Module.require (internal/modules/cjs/loader.js:1025:19)
        at require (internal/modules/cjs/helpers.js:72:18)
```

This appears to be due to the Pulumi 2.0 transition, and has probably been broken since the example was updated to use 2.0 in on 4/20/2020.

PR #774, opened 2020-08-24
Fixup the GKE Rails Applicaton to adhere to Pulumi 2.0 code style

stack72:
Fixes: #773


joeduffy:
LGTM, merging so I can get back to the customer. Thanks for the quick fix, @stack72 !

webmaeistro:
:D Joe! you the man! i believe this will do the trick

PR #801, opened 2020-09-22
Update Python examples to use classes

### LINKED ISSUE #700: Update Python examples to use classes
Update Python examples use classes rather than dicts for nested inputs/outputs.

Update any copyright years as part of doing this.

Part of https://github.com/pulumi/pulumi/issues/3771.

justinvp:
Mostly mechanical changes from dicts for nested data to new classes, and minor improvements to `README.md` and `Pulumi.yaml` as I went.

Fixes #700

### DIFF
--- a/aws-py-appsync/__main__.py
+++ b/aws-py-appsync/__main__.py
@@ -7,10 +7,10 @@ from pulumi_aws import dynamodb, iam, appsync
 table = dynamodb.Table(
     "tenants",
     hash_key="id",
-    attributes=[{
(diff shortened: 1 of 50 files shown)


stack72:
@justinvp this is brilliant - I am happy that if the tests pass here then we should ship this 👍 

If there are any small things that need updated then we can do that out of band

justinvp:
Thanks, @stack72! I think the tests look good.

stack72 (approved):

PR #802, opened 2020-09-23
Update the webhook example to use Slack apps instead of legacy integration

leezen:
Previously, the webhook example was using the old Slack integration style. This PR updates it to use the newer app-style integration for the incoming webhook.

### DIFF
--- a/aws-ts-pulumi-webhooks/Pulumi.yaml
+++ b/aws-ts-pulumi-webhooks/Pulumi.yaml
@@ -1,3 +1,13 @@
 name: aws-ts-pulumi-webhooks
 runtime: nodejs
 description: Lambda app that handles Pulumi webhook notifications.
+template:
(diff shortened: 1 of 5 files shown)


cnunciato (commented):
Looks good, but we should probably update the README to match these changes, and ideally (since we're in here) add a couple of lines to Pulumi.yaml to define a template with the config params expected by this app, as it'd make the Deploy button work more completely.

Maybe outside the scope of this PR, but part of me would also like to see the webhook URL marked as a secret, given how it works. It'd complicate the program because of the closed-over value, so maybe a comment and a link to some docs (if we have any) could suffice for it. Just figured I'd mention it as something to consider, though.

jkisk (approved):
This looks good, I like Christian's suggestion to add the pulumi.yaml too.

PR #806, opened 2020-10-04
Compare and contrast azure-py-vdc and azureng-py-vdc

jamesianberry:
This includes both an update to azure-py-virtual-data-center and a new azureng-py-virtual-data-center.

The two examples are interoperable - you can peer a stack in one with a stack in the other. Both implement the latest Azure Firewall features including forced tunnelling - although the current azure provider can't have additional properties.

Code has been refactored so that both examples are directly comparable - examine vdc.py to see the differences between providers. For example, some parameter names have changed e.g. public_ip_address is now publicIPAddress. Differences in other files are to do with the lack of auto-naming in azure-nextgen.

Code has been optimised to overcome contention issues that show up in the activity log, by moving things around and adding depends_on statements.

These examples demonstrate custom routing that avoids a common pitfall, where traffic already headed to one member of the Azure Firewall VM Scale Set is redirected to the private IP address - causing looping and dropped traffic.


mikhailshilkov:
This looks great @jamesianberry, amazing job!

One high-level question: what do you think about using argument classes instead of untyped dictionaries? (see [this example](https://github.com/pulumi/examples/blob/master/azure-nextgen-py-aci/__main__.py#L20)) They are a relatively new addition but I find them very valuable in terms of safety and IDE support.

mikhailshilkov (approved):

PR #807, opened 2020-10-05
Add examples of mocks to unit tests

mikhailshilkov:
- For all four runtimes, add an example of how to mock resources and invokes
- Add an invoke to each program
- Fix some issues caused by a previous upgrade to AWS v3

Resolves https://github.com/pulumi/pulumi/issues/5295


justinvp (approved):
LGTM

Issue #804, opened 2020-09-30
azure-ts-dynamicresource example fails to enable https


devployment:
Following [the example](https://www.pulumi.com/docs/tutorials/azure/azure-ts-dynamicresource/) I get an error if I try to enable https for the custom domain.

```
error: The resource format is invalid.
```

Creating the custom domain works, without https enabled. As soon I try to enable it either on create, or as a change I get the resource format error.

If I use `az cdn custom-domain enable-https` from the Azure CLI directly on an available custom domain it works with the same same parameter given to `cdnClient.customDomains.enableCustomHttps`

I guess it's related to https://github.com/Azure/azure-sdk-for-js in general and maybe [somehow to this](https://github.com/Azure/azure-cli/issues/12273) from the CLI. But maybe somebody in here has another idea.

ChrisKlug:
The reason is that version 5 of @azure/arm-cdn requires an extra parameter to be passed in to enableCustomHttps. A working version looks like this

```
await cdnClient.customDomains.enableCustomHttps(
    inputs.resourceGroupName,
    inputs.profileName,
    inputs.endpointName,
    this.name,
    {
        customDomainHttpsParameters: {
            certificateSource: "Cdn",
            certificateSourceParameters: {
                certificateType: "Dedicated"
            },
            protocolType: "ServerNameIndication"
        }
    });
```

I'm not sure how Pulumi handles versions for @azure/XXX dependencies. But I have created a PR that fixes the issue when using version 5. Not sure what happens when using 4.2 though... But until that potentially gets accepted, the above code should fix it!

PR #812, opened 2020-10-12
Fixing problem with adding custom HTTPS to CDN

ChrisKlug:
Fix to #804

The latest version (5.0) of @azure/arm-cdn requires an extra parameter to be passed in to enableCustomHttps. The current version does not provide this parameter, and fails with error: The resource format is invalid.

To fix this, the CDNCustomDomainResourceProvider has been updated to provide some default values to this parameter. And the @azure/arm-cdn has been bumped to 5.0 as this is the latest version.


praneetloke (approved):
@ChrisKlug thank you so much for taking the time to update this example. Much appreciated!

ChrisKlug:
Just a thing to note here is that I got a PR that fixes this approved for the Azure JS SDK as well. It hasn't been merged yet though. And, this fix will work fine with both the unfixed SDK and the fixed. So it shouldn't be a problem to merge it

PR #816, opened 2020-10-23
Document the allowed chars in the session name

MattBlack85:


leezen (approved):
Thanks for the submission

PR #817, opened 2020-10-23
add sparse checkout instructions to README

spara:
there are a lot of examples and sometimes you only want that one example


mikhailshilkov (approved):
LGTM, thank you!

PR #834, opened 2020-11-16
Moving away from deprecated GHA add-path

stack72:


stack72:
Self merging as this is only a mechanical GHA change for syntax - no test functionality changed and we have a green build!

PR #836, opened 2020-11-18
Update aws-go-lambda README

leezen:


komalali (approved):

aws-go-lambda/README.md
```diff
@@ -60,20 +61,20 @@ After cloning this repo, run these commands from the working directory:
...
 	```bash
 	aws lambda invoke \
 	--function-name $(pulumi stack output lambda) \
 	--region $(pulumi config get aws:region) \
-	--payload '"foo"' \
+	--payload 'ImZvbyI=' \
 	output.json
 
 	cat output.json # view the output file with your tool of choice
 	# "FOO"
```
komalali, line 74: The output is actually "foo".
```suggestion
	# "foo"
```

aws-go-lambda/README.md
```diff
@@ -1,5 +1,5 @@
 # AWS Golang Lambda
-This example creates a lambda that does a simple `ToUpper` on the string input and returns it.
+This example creates an AWS Lambda function that does a simple `ToUpper` on the string input and returns it.
```
komalali, line 2: It doesn't do a `ToUpper`. It seems to decode base64.

komalali (changes requested):

komalali (approved):

PR #837, opened 2020-11-18
Update aws-cs-lambda README

komalali:
Inspired by https://github.com/pulumi/examples/pull/836

Fixes up this page: https://www.pulumi.com/docs/tutorials/aws/aws-cs-lambda/


leezen (approved):
Thanks for proactively finding this!

PR #839, opened 2020-11-19
Fixed issue where provisioner example changeToken doesn't work

### LINKED ISSUE #838: Provisioners example - changeToken doesn't work
Looking at the file [provisioner.ts]( https://github.com/pulumi/examples/blob/master/aws-ts-ec2-provisioners/provisioners/provisioner.ts)

It seems to offer some nice functionality such that a change token can be used; this is very useful for my application. Unfortunately, it appears to be bugged, and the change token doesn't work. 

I've made a simple fix on my end which seems to work; PR incoming.

JustASquid:
Fixes #838.

### DIFF
--- a/aws-ts-ec2-provisioners/provisioners/provisioner.ts
+++ b/aws-ts-ec2-provisioners/provisioners/provisioner.ts
@@ -39,7 +39,7 @@ export class Provisioner<T, U> extends pulumi.dynamic.Resource {
                 return { id: uuid.v4(), outs: inputs };
             },
         };
-        super(provider, name, { dep: props.dep, result: null }, opts);
+        super(provider, name, { dep: props.dep, changeToken: props.changeToken, result: null }, opts);
     }
 }
 


stack72:
Hi @JustASquid 

this is fantastic! Before this change I can see (with a change to myapp.conf)

```
▶ pulumi up
Previewing update (dev)


     Type                 Name                         Plan
     pulumi:pulumi:Stack  aws-ts-ec2-provisioners-dev

Resources:
    8 unchanged
```

But afterwards I get:

```
▶ pulumi up
Previewing update (dev)

     Type                                  Name                         Plan        Info
     pulumi:pulumi:Stack                   aws-ts-ec2-provisioners-dev
     ├─ pulumi:provisioners:CopyFile       config
 +-  │  └─ pulumi-nodejs:dynamic:Resource  config-provisioner           replace     [diff: +changeToken]
     └─ pulumi:provisioners:RemoteExec     cat-config
 +-     └─ pulumi-nodejs:dynamic:Resource  cat-config-provisioner       replace     [diff: +changeToken]

Outputs:
  - catConfigStdout: "[test]\nx = 42\n"

Resources:
    +-2 to replace
    6 unchanged
```

Thank you so much for this!

Paul

PR #858, opened 2020-12-12
Do more complete testing

lukehoban:
We want to make sure our test coverage for examples tests the full lifecycle developers will experience with our examples, as well as gain coverage for a broad set of core Pulumi experience.  The previous settings were significantly too conservative - meaning we were missing coverage of three traditionally sensitive areas of the product:
1. Initial preview
2. Refresh
3. Empty preview

The few tests that cannot correctly pass (3) are exempted temporarily and individually from that one step - and https://github.com/pulumi/examples/issues/859 is tracking undoing this to ensure full coverage.

### DIFF
--- a/misc/test/examples_test.go
+++ b/misc/test/examples_test.go
@@ -1154,6 +1154,10 @@ func TestAccDigitalOceanPyK8s(t *testing.T) {
 					return assert.Contains(t, body, "Welcome to nginx!")
 				})
 			},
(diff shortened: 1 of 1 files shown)


leezen (approved):

lukehoban:
Confirmed that all 13 failing tests are due to https://github.com/pulumi/pulumi/issues/5933, which are expected failures that should be fixed in `v2.15.6`.

Issue #782, opened 2020-09-07
Dynamic resources / Provisioners / CopyFile and RemoteExec issues / TS & Python


szymonbr:
Recently I've used Python provisioner module for setting up an application settings at some set of Azure VMs. And while making some config adjustments I found that the configs are not updated at all. There're exactly the same like after initial spawning of the VMs. 

Hence I've tested updating an app configuration with pulumi dynamic resources again with using examples from this repo: 

- aws-py-ec2-provisioners
- aws-ts-ec2-provisioners

From my test it seems that both TS i Python provisioner modules are not reliable and don't work correctly. 

TS provisioners are not refreshing resources after a file modification (concerns `CopyFile`) as well as a command change (concerns `RemoteExec`).
Whereas in case of using Python provisioners `CopyFile` doesn't work after a file modification at all but  `RemoteExec` works slightly different. It refreshes dynamic resource every time I run `pulumi up` even when nothing is changed. And at least it works when I make a change in a command list. 

Maybe someone have any idea how to fix it, especially for Python Provisioner module?

swiftugandan:
Am seeing similar issues too

szymonbr:
I've found that it's not complete solution but rather some kind of preview. 
Unless I'm wrong (I newby in Python) both `CopyFileProvider `and `RemoteExecProvider `classes have only got `on_create` method each which is not applicable to make any update of such dynamic resources. To apply changes dynamic resources have to be recreated. It would be great if somebody could make an update of those classes to add `on_change` methods.

AlexanderSarson:
I got the same issue

leezen:
The `aws-ts-ec2-provisioners` example was fixed in https://github.com/pulumi/examples/pull/839 and you should see a replacement when updating the file. There is an issue with the Python provisioner in that it's not recording anything in the state to track if the file itself changes. Changes to the connection should cause a replacement. I have a PR to fix the Python issue.

PR #864, opened 2020-12-22
Include file hash in the resource state for the Python provisioner example

leezen:
Fixes https://github.com/pulumi/examples/issues/782


komalali (approved):

leezen:
@stack72 Thanks for merging on my behalf. I missed this getting approved somehow.

PR #868, opened 2020-12-22
Fix tests

lukehoban:
AllEmptyChanges on three tests that show changes on empty updates (until these can be investigated and hopefully fixed in the examples or the product).


lukehoban:
Tests are green except for Kubernetes which seems like an independent issue. Opened #869 to track that.

PR #870, opened 2021-01-04
Add a flag to command in Readme for AKS example

mashun4ek:
Add a flag `--show-secrets` to ensure that kubeconfig file is populated with necessary information.


stack72:
Hi @mashun4ek 

Thank you so much for your PR here - did you find that this was actually needed? I've never needed to set this so I am confused and worried there may be a regression here

Pal

mashun4ek:
Hey @stack72 👋

Hmm, after I run the command `pulumi stack output kubeconfig` the output is: [secret]. So I had to use `--show-secrets` flag to see my kubeconfig. 
I just started using pulumi and azure, so it might be an issue on my side, disregard my pr in that case please :)

stack72 (approved):

PR #882, opened 2021-01-23
Split up the acceptance tests to be able to use build tags

stack72:
This will allow us to run all tests as part of PRs and crons
and also to run a specific set of them for things like smoke testing
providers


mikhailshilkov (approved):
LGTM

stack72:
Thanks @mikhailshilkov 

just a FYI, these failures are also in master so they are unrelated - I am looking at them now

PR #885, opened 2021-01-26
Add `virtualenv` option to Python examples

### LINKED ISSUE #760: Update all Python examples to have the `virtualenv` option already specified in Pulumi.yaml
And update README instructions as appropriate.

When running `pulumi preview` or `pulumi up`, if the virtual environment directory doesn't exist, an error is returned with the commands to run to create the virtual environment. But with https://github.com/pulumi/pulumi/issues/5164, we would handle the virtual environment creation automatically/transparently so the user would not have to run any additional commands.

lukehoban:
Use the automatically managed `virtualenv: venv` support in all Python examples.

See https://www.pulumi.com/docs/intro/languages/python/#virtual-environments.

Fixes #760.

### DIFF
--- a/aws-django-voting-app/Pulumi.yaml
+++ b/aws-django-voting-app/Pulumi.yaml
@@ -1,5 +1,8 @@
 name: voting-app
-runtime: python
+runtime:
+  name: python
+  options:
+    virtualenv: venv
 description: Voting app that uses containers
 template:
   config:
--- a/aws-django-voting-app/README.md
+++ b/aws-django-voting-app/README.md
@@ -39,16 +39,6 @@ The example shows how easy it is to deploy containers into production and to con
     $ pulumi config set django-secret-key <VALUE> --secret
     ```
 
-1. Create a Python virtualenv, activate it, and install dependencies:
- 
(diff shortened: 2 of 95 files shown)


lukehoban:
@justinvp You noted in #760 that we should also remove the README sections that suggest running virtualenv commands. Curious your thoughts on whether it's still a good idea to remove those as well?

stack72:
@lukehoban we need to update the go test code as well

```
error: failed to discover plugin requirements: The 'virtualenv' option in Pulumi.yaml is set to "venv", but "/tmp/p-it-fv-az148-2-digitaloce-5da82d26-517438308/venv" doesn't exist; run the following commands to create the virtual environment and install dependencies into it:
```

we need to set `UseAutomaticVirtualEnv: true` in each of the Test funcs that are changed (and relate to a test func)

justinvp:
> @justinvp You noted in #760 that we should also remove the README sections that suggest running virtualenv commands. Curious your thoughts on whether it's still a good idea to remove those as well?

Yes, we should edit those out of the READMEs.

> we need to set `UseAutomaticVirtualEnv: true` in each of the Test funcs that are changed (and relate to a test func)

Yep.

@lukehoban, I can take care of these, if you'd like. I'll push to your branch.

stack72 (approved):
This is awesome! Thank you both!!

PR #893, opened 2021-01-27
Mitch/aws py wordpress fargate rds

MitchellGerdisch:
Python example that deploys WordPress in ECS with an RDS backend.


stack72:
@MitchellGerdisch this is fantastic - I just took this for a spin and it worked well. The docs are very descriptive and easy to follow! I especially like that it demonstrates the use of ComponentResources

Going to merge this and we can follow up with any changes needed afterwards

PR #905, opened 2021-02-08
New example: Secret rotation with one set of credentials

MisinformedDNA:
I thought this would make a good example. Thoughts?

I'm also working though secret rotation with two sets of credentials.


azure-cs-credential-rotation-one-set/Azure.CredentialRotation.OneSet.csproj
```diff
@@ -0,0 +1,19 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Pulumi.Azure" Version="3.*" />
+    <PackageReference Include="Pulumi.AzureNextGen" Version="0.4.0" />
```
mikhailshilkov, line 11: Consider setting it to `0.*` or `0.6.0`.

azure-cs-credential-rotation-one-set/Azure.CredentialRotation.OneSet.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Pulumi.Azure" Version="3.*" />
```
mikhailshilkov, line 10: Why is this one needed?

azure-cs-credential-rotation-one-set/AppStack.cs
```diff
@@ -0,0 +1,315 @@
...
+                Tags = { { "CredentialId", sqlAdminLogin }, { "ProviderAddress", sqlServer.Id }, { "ValidityPeriodDays", "1" } },
+                Properties = new SecretPropertiesArgs
+                {
+                    Value = sqlAdminPassword,
+                    Attributes = new SecretAttributesArgs { Expires = (int)expiresAt },
+                },
+            },
+            new CustomResourceOptions { DependsOn = { eventSubscription } }); 
+
+        #region WebApp
```
mikhailshilkov, line 248: Nit: we prefer not to use regions in our code

azure-cs-credential-rotation-one-set/AppStack.cs
```diff
@@ -0,0 +1,315 @@
...
+internal class AppStack : Stack
+{
+    [Output]
+    public Output<string> WebAppEndpoint { get; set; }
+
+    public AppStack()
+    {
+        var config = new Config();
+        var resourceGroupName = config.Require("resourceGroupName");
+        var resourceGroup = Output.Create(GetResourceGroup.InvokeAsync(new GetResourceGroupArgs
```
mikhailshilkov, line 28: Any objection to creating a new resource group here? We try to make our examples self-contained.

mikhailshilkov (commented):
This looks great, thank you @MisinformedDNA! Could you please rename it to `azure-nextgen-cs-...`?

azure-cs-credential-rotation-one-set/AppStack.cs
MisinformedDNA: No objection, but this is precisely the issue I've been trying to get at. If someone using the arm2pulumi generator wants their stack to be self-contained and use `new ResourceGroup`, like the rest of the examples, then they have to fix all the resource group references, which isn't always a quick task with bigger ARM templates. Switching to `Output.Create(new ResourceGroup)` is an easy fix, but doesn't match the other examples.

azure-cs-credential-rotation-one-set/Azure.CredentialRotation.OneSet.csproj
MisinformedDNA: `FirewallRule` and `AccessPolicy`

azure-cs-credential-rotation-one-set/Azure.CredentialRotation.OneSet.csproj
MisinformedDNA: Fixed

azure-cs-credential-rotation-one-set/AppStack.cs
MisinformedDNA: Removed.

azure-cs-credential-rotation-one-set/AppStack.cs
MisinformedDNA: I used `new ResourceGroup`.

Examples.sln
```diff
@@ -51,6 +51,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "UnitTesting", "testing-unit
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Azure.Functions.LinuxApp", "azure-cs-functions-linux-app\Azure.Functions.LinuxApp.csproj", "{B3739A06-5343-4990-A48B-E5BEFAC5678A}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Azure.CredentialRotation.OneSet", "azure-nextgen-cs-credential-rotation-one-set\Azure.CredentialRotation.OneSet.csproj", "{34988495-6782-4AE3-9224-DE26A8FA1C61}"
```
MisinformedDNA, line 54: Did you want this added to the solution? None of the other NextGen projects are included. Also, should the project name be renamed with AzureNextGen?

azure-cs-credential-rotation-one-set/AppStack.cs
viveklak: > No objection, but this is precisely the issue I've been trying to get at. If someone using the arm2pulumi generator wants their stack to be self-contained and use `new ResourceGroup`, like the rest of the examples, then they have to fix all the resource group references, which isn't always a quick task with bigger ARM templates. Switching to `Output.Create(new ResourceGroup)` is an easy fix, but doesn't match the other examples.

I cut https://github.com/pulumi/arm2pulumi/issues/24 to track this. Let me know your thoughts there.

azure-cs-credential-rotation-one-set/Azure.CredentialRotation.OneSet.csproj
mikhailshilkov: Anything wrong with the NextGen `FirewallRule`? https://www.pulumi.com/docs/reference/pkg/azure-nextgen/sql/firewallrule/

I know we have an issue tracking the access policy. But you could define the `WebApp` resource before the `KeyVault` resource and then define web app's settings as a separate resource. https://www.pulumi.com/docs/reference/pkg/azure-nextgen/web/webapplicationsettings/

Examples.sln
mikhailshilkov: Yes, I think we should add all NextGen projects eventually

azure-cs-credential-rotation-one-set/Azure.CredentialRotation.OneSet.csproj
MisinformedDNA: Sorry that I ask so many questions, but does `WebApplicationSettings` get mapped from [WebApps_UpdateApplicationSettings](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/web/resource-manager/Microsoft.Web/stable/2020-10-01/WebApps.json#L1012)? If so, shouldn't the class be named `WebAppApplicationSettings`? It sounds a bit repetitive, but it loses consistency.

mikhailshilkov: Hmm, it does, and it should. Mind opening an issue and I'll take a look at what goes wrong?

MisinformedDNA: https://github.com/pulumi/pulumi-azure-nextgen/issues/282

MisinformedDNA: Done.

azure-nextgen-cs-credential-rotation-one-set/AppStack.cs
```diff
@@ -0,0 +1,304 @@
...
+    {
+        var config = new Config();
+        var resourceGroupName = config.Get("resourceGroupName") ?? "rotatesecretoneset";
+        var resourceNamePrefix = config.Get("resourceNamePrefix") ?? resourceGroupName;
+        var location = config.Get("location") ?? "EastUS";
+
+        var resourceGroup = new ResourceGroup("resourceGroup", new ResourceGroupArgs { ResourceGroupName = resourceGroupName, Location = location });
+
+        var sqlAdminLogin = config.Get("sqlAdminLogin") ?? "sqlAdmin";
+        var sqlAdminPassword = Guid.NewGuid().ToString();
```
mikhailshilkov, line 35: This will change on every run, so I expect you see a diff on every `pulumi up`. It's better to use the random provider for this (or just hard-code for simplicity).

azure-nextgen-cs-credential-rotation-one-set/AppStack.cs
```diff
@@ -0,0 +1,304 @@
...
+                ServerName = sqlServer.Name,
+                ResourceGroupName = resourceGroup.Name,
+                StartIpAddress = "0.0.0.0",
+                EndIpAddress = "0.0.0.0",
+            });
+
+        var clientConfig = Output.Create(GetClientConfig.InvokeAsync());
+        var tenantId = clientConfig.Apply(c => c.TenantId);
+
+        var storageAccountName = $"st{resourceNamePrefix}";
```
mikhailshilkov, line 59: Nit: if it's a prefix, why append it?

azure-nextgen-cs-credential-rotation-one-set/AppStack.cs
```diff
@@ -0,0 +1,304 @@
...
+        var storageAccount = new StorageAccount("storageAccount", new StorageAccountArgs
+        {
+            AccountName = storageAccountName,
+            Kind = "Storage",
+            Location = location,
```
mikhailshilkov, line 64: If you update to 0.6.1, the location shouldn't be a required parameter anymore, similar to the classic Azure provider

mikhailshilkov (approved):
LGTM! Left several minor remarks.

azure-nextgen-cs-credential-rotation-one-set/AppStack.cs
MisinformedDNA: Yeah, I did that because I thought it would be more secure if no one ever knew what the password was. But the password changes within a minute anyway, so maybe it's over the top. :)

azure-nextgen-cs-credential-rotation-one-set/AppStack.cs
MisinformedDNA: Fixed

azure-nextgen-cs-credential-rotation-one-set/AppStack.cs
MisinformedDNA: Fixed

azure-nextgen-cs-credential-rotation-one-set/AppStack.cs
MisinformedDNA: Done.

mikhailshilkov (approved):
LGTM, thank you!

PR #932, opened 2021-03-04
Add support for AzureWebJobsStorage azure-cs-functions

### LINKED ISSUE #926: AzureNative function doesn't create as the classic example
## Expected behavior
AzureNative c# functions example should be the same as the Classic

## Current behavior
The Azure Function c# example creates a function using the WebApp type, however this omits the AzureWebJobsStorage elements of an Azure function which is vitally important to a lot of functions usecases (host keys for one) and therefore causes weird issues for users.

## Steps to reproduce
Run in the example and try to create function keys in the portal.

## Context (Environment)
I'm having to now workout how to extract account keys from a `StorageAccount` resource in order to add the account keys.

## Affected feature
Creating Azure Functions through Pulumi

viveklak:
Fixes #926

### DIFF
--- a/azure-cs-functions/FunctionsStack.cs
+++ b/azure-cs-functions/FunctionsStack.cs
@@ -79,6 +79,10 @@ class FunctionsStack : Stack
                 AppSettings = new[]
                 {
                     new NameValuePairArgs{
+                        Name = "AzureWebJobsStorage",
+                        Value = GetConnectionString(resourceGroup.Name, storageAccount.Name),
+                    },
+                    new NameValuePairArgs{
                         Name = "runtime",
                         Value = "python",
                     },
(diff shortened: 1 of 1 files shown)


azure-cs-functions/FunctionsStack.cs
```diff
@@ -78,6 +78,10 @@ public FunctionsStack()
             {
                 AppSettings = new[]
                 {
+                    new NameValuePairArgs{
+                        Name = "AzureWebJobsStorage",
+                        Value = GetConnectionString(resourceGroup.Name, storageAccount.Name),
+                    },
```
mikhailshilkov, line 84: Unrelated to this change but I think it would be good to use a separate `WebAppApplicationSettings` resource which is very useful in practice

mikhailshilkov (approved):
TY!

azure-cs-functions/FunctionsStack.cs
viveklak: Agreed. That said, we have some coverage here now: https://github.com/pulumi/examples/blob/master/azure-cs-credential-rotation-one-set/AppStack.cs#L249.

PR #937, opened 2021-03-09
TypeScript version of webserver example for language parity

davidwrede:


aws-ts-webserver/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// (optional) create a simple web server using the startup script for the instance
+const userData =
+`#!/bin/bash
+echo "Hello, World!" > index.html
+nohup python -m SimpleHTTPServer 80 &`;
+
+const server = new aws.ec2.Instance("web-server-www", {
+    tags: { "Name": "web-server-www" },
+    instanceType: size,
```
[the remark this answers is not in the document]
davidwrede: `pulumi up` run successfully, but you're right, I converted this one a bit too quickly from JS and didn't make it TypeScript native. Will set this to draft and revise.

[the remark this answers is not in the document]
davidwrede: I spot-checked our docs and I couldn't find another example of how I should handle this in a TypeScript way, so I'm leaving it as is (since it works). If there's a more TypeScript way of handling this, just let me know and I'll change it.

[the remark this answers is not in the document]
leezen: This is implemented as a relaxed enum, so it'll take a string. For our examples, I think it'd be preferable to show this with an enum: `aws.ec2.InstanceType.T2_Small`

README.md
```diff
@@ -94,6 +94,7 @@ Example   | Description |
 [Twitter](aws-ts-twitter-athena) | Query Twitter every 2 minutes, store the results in S3, and set up an Athena table and query.
 [URL Shortener](aws-ts-url-shortener-cache-http) | Create a serverless URL shortener that uses high-level components.
 [Voting App](aws-ts-voting-app) | Create a simple voting app using Redis and Python Flask.
+[Web Server](aws-ts-webserver) | Deploy an EC2 Virtual machine running a Python web server.
```
leezen, line 97:
```suggestion
[Web Server](aws-ts-webserver) | Deploy an EC2 Virtual machine using TypeScript to run a Python web server.
```

leezen: Not sure if the suggestion above is necessary, but given we're mixing languages, it might be useful to clarify.

aws-ts-webserver/README.md
```diff
@@ -0,0 +1,61 @@
...
+    ```
+
+4. After a couple minutes, your VM will be ready, and two stack outputs are printed:
+
+    ```bash
+    $ pulumi stack output
+    Current stack outputs (2):
+    OUTPUT          VALUE
+    publicIp        53.40.227.82
+    ```
```
leezen, line 45: This only shows one of the outputs?

aws-ts-webserver/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+5. Thanks to the security group making port 80 accessible to the 0.0.0.0/0 CIDR block, we can curl it:
```
leezen, line 47:
```suggestion
5. Thanks to the security group making port 80 accessible to the 0.0.0.0/0 CIDR block (all addresses), we can curl it:
```

aws-ts-webserver/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import * as pulumi from "@pulumi/pulumi";
+
+// Get the id for the latest Amazon Linux AMI
+const ami = aws.ec2.getAmi({
+    filters: [
+        { name: "name", values: ["amzn-ami-hvm-*-x86_64-ebs"] },
+    ],
+    owners: ["137112412989"], // Amazon
+    mostRecent: true,
+}, { async: true }).then(result => result.id);
```
leezen, line 13:
```suggestion
}).then(result => result.id);
```

We no longer need `async: true` post-2.0

leezen (approved):
Some minor nits/suggestions

Issue #938, opened 2021-03-10
classic-azure-ts-aks-helm failing during pulumi up


t0yv0:
Failing to `pulumi up` in classic-azure-ts-aks-helm.

## Expected behavior

Helm chart installed and stood up in the K8S cluster.

## Current behavior

Errors as follows:

```
    STACK_TRACE:
    Error:
        at Object.debuggablePromise (/Users/anton/examples/classic-azure-ts-aks-helm/node_modules/@pulumi/pulumi/runtime/debuggable.js:69:75)
        at /Users/anton/examples/classic-azure-ts-aks-helm/node_modules/@pulumi/pulumi/runtime/invoke.js:126:45
        at Generator.next (<anonymous>)
        at fulfilled (/Users/anton/examples/classic-azure-ts-aks-helm/node_modules/@pulumi/pulumi/runtime/invoke.js:18:58)
        at processTicksAndRejections (internal/process/task_queues.js:97:5)
    unhandled rejection: CONTEXT(968): Invoking function: tok=kubernetes:helm:template asynchronously
    STACK_TRACE:
    Error:
        at Object.debuggablePromise (/Users/anton/examples/classic-azure-ts-aks-helm/node_modules/@pulumi/pulumi/runtime/debuggable.js:69:75)
        at /Users/anton/examples/classic-azure-ts-aks-helm/node_modules/@pulumi/pulumi/runtime/invoke.js:126:45
        at Generator.next (<anonymous>)
        at fulfilled (/Users/anton/examples/classic-azure-ts-aks-helm/node_modules/@pulumi/pulumi/runtime/invoke.js:18:58)
        at processTicksAndRejections (internal/process/task_queues.js:97:5)

    error: Running program '/Users/anton/examples/classic-azure-ts-aks-helm' failed with an unhandled exception:
    Error: invocation of kubernetes:helm:template returned an error: failed to generate YAML for specified Helm chart:
        failed to pull chart: chart "bitnami/apache" version "1.0.0" not found in
        https://charts.bitnami.com/bitnami repository
        at Object.callback (/Users/anton/examples/classic-azure-ts-aks-helm/node_modules/@pulumi/pulumi/runtime/invoke.js:139:33)
        at Object.onReceiveStatus (/Users/anton/examples/classic-azure-ts-aks-helm/node_modules/@grpc/grpc-js/src/client.ts:334:26)
[...]

PR #940, opened 2021-03-11
Update the classic azure TypeScript AKS helm example

t0yv0:
Got this working finally.

Doc additions may be getting a bit verbose. I'm open to replacing with links to "getting started" with Pulumi on TypeScript (tsc, npm install), and Azure (az login) as the example may not be the right place for that.


t0yv0:
Resolves https://github.com/pulumi/examples/issues/938

mikhailshilkov (approved):

PR #941, opened 2021-03-11
Add azure-ts-aks-helm

t0yv0:
Adds a port of the classic-azure helm chart example to the new native provider.


mikhailshilkov (approved):
Looks great! Do you mind porting to other languages?

t0yv0:
Sure! WIll do. I was also looking at how to add it to the test suite so that nightlies verify that it keeps working.

PR #942, opened 2021-03-11
Apply gofmt

t0yv0:
```bash

$ go version
go version go1.16 darwin/amd64

$ gofmt -w $(git ls-files | grep '.go$')

```


t0yv0:
No huge deal but @stack72 we could make a CI thingy that rejects if go file is not gofmt-'d. I can look into it later not worked with GH actions before :)

mikhailshilkov (approved):
👍

PR #944, opened 2021-03-16
Adding TypeScript version of AWS S3 Website example

davidwrede:
Here's a TypeScript version of creating a static website on AWS.

Note: This version explicitly creates a BucketPolicy resource, which isn't how it's done amongst the other versions of this example. It feels this is a better way of handling things (versus how it's handled in the JS and Python versions), and if folks agree, I can update the other versions to match this style.


aws-ts-s3-folder/index.ts
```diff
@@ -0,0 +1,48 @@
...
+        bucket: siteBucket,                               // reference the s3.Bucket object
+        source: new pulumi.asset.FileAsset(filePath),     // use FileAsset to point to a file
+        contentType: mime.getType(filePath) || undefined, // set the MIME type of the file
+    });
+}
+
+// Set the access policy for the bucket so all objects are readable
+const bucketPolicy = new aws.s3.BucketPolicy("bucketPolicy", {
+    bucket: siteBucket.id, // refer to the bucket created earlier
+    policy: pulumi.all([siteBucket.arn]).apply((bucketArn) => JSON.stringify({
```
komalali, line 31:
```suggestion
    policy: siteBucket.arn.apply((bucketArn) => JSON.stringify({
```

stack72 (approved):

PR #951, opened 2021-03-22
azure-py-aks-helm

t0yv0:
Straightforward port of the TypeScript example, no big surprises here. 

- Needs https://github.com/pulumi/pulumi/pull/6599 fix to manage venvs correctly

- Unrelated? example TestAccAwsPyAssumeRole test failure


azure-py-aks-helm/__main__.py
```diff
@@ -0,0 +1,27 @@
+"""Provisions Apache via a Helm chart onto an AKS cluster created in
```
mikhailshilkov, line 1: Copyright in all files

mikhailshilkov (approved):
LGTM

PR #953, opened 2021-03-23
Consistently use ApplyT

### LINKED ISSUE #952: Update Go examples to use ApplyT instead of Apply<TypeName>
Related: https://github.com/pulumi/pulumi/issues/6592

lukehoban:
25 of the 30 uses of `.Apply*` functions in Pulumi Go examples were using `ApplyT`.  Update the remaning 5 to also use `ApplyT`.

Fixes #952.

### DIFF
--- a/aws-go-fargate/main.go
+++ b/aws-go-fargate/main.go
@@ -118,7 +118,7 @@ func main() {
 			return err
 		}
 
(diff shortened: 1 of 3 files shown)


t0yv0 (approved):
LGTM

PR #959, opened 2021-04-03
Ported and created Azure Native AKS multi-cluster examples

davidwrede:
Some of these are ports of the classic Azure versions while the C# one is new, but based off of the others. 

Note: I've got a couple of weird spacing issues that need to be fixed, but they shouldn't block a code review.


azure-go-aks-multicluster/go.mod
```diff
@@ -0,0 +1,9 @@
+module azure-go-aks-multicluster
+
+go 1.15
+
+require (
+	github.com/pulumi/pulumi-azure-native/sdk v0.7.1
```
viveklak, line 6:
```suggestion
	github.com/pulumi/pulumi-azure-native/sdk v0.8.0
```

azure-go-aks-multicluster/go.mod
```diff
@@ -0,0 +1,9 @@
...
+	github.com/pulumi/pulumi-azuread/sdk/v3 v3.5.1
+	github.com/pulumi/pulumi/sdk/v2 v2.21.2
```
viveklak, line 8:
```suggestion
	github.com/pulumi/pulumi/sdk/v2 v2.24.1
```

Sadly Go will pin these versions and we are still looking for a good automated means to keep versions up-to-date.

azure-go-aks-multicluster/go.mod
```diff
@@ -0,0 +1,9 @@
+module azure-go-aks-multicluster
+
+go 1.15
```
viveklak, line 3:
```suggestion
go 1.16
```
You might have to regen the go.sum etc. but recommend bumping to 1.16.

viveklak (approved):
Thanks for this! Minor nits on the Go impl but otherwise LGTM.

PR #963, opened 2021-04-12
(bug)use map string instead of map

ranjib:


jaxxstorm:
Thankyou for the fix!

lukehoban:
Does this code compile?  I would have expected this to need to be `pulumi.StringMap` (not `MapString`: https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v2/go/pulumi#StringMap

ranjib:
Yes. Without this I was hitting compilation errors

PR #980, opened 2021-04-22
Azure C# AKS + Cosmos DB example

mikhailshilkov:
Port of https://github.com/pulumi/examples/tree/master/classic-azure-ts-aks-mean

I used it for a demo, so decided to save it for the future. Very similar to an existing C# Helm example - not sure if we want to keep both or replace it?


t0yv0 (approved):
Sweet! Let's include for now,  I don't know what the tradeoffs are of having too many examples. We can do a pass and trim them up later, using engagement data or something like that to rank them.

PR #977, opened 2021-04-21
Example of integrating Azure SDK to TypeScript

mikhailshilkov:
Part of https://github.com/pulumi/pulumi-azure-native/issues/601

PR #981, opened 2021-04-22
C# integrate Azure API

mikhailshilkov:
Part of https://github.com/pulumi/pulumi-azure-native/issues/601

C# version of https://github.com/pulumi/examples/pull/977

It turns out that the .NET SDK for this function is old and doesn't support .NET Core, so I had to make a raw HTTP request to achieve the same. We can combine it with [this example of an SDK call](https://github.com/pulumi/pulumi-azure-native/blob/3e3cf5fefca7a4580aef0fc313635ff76c6d0428/examples/cs-simple/MyStack.cs#L71) which should be enough.


mikhailshilkov:
@viveklak if you have any spare time, I'd really appreciate if you could try to put a Python and/or Go version of this together.

Issue #960, opened 2021-04-08
Deprecated addon in Azure Python AKS example


spara:
https://github.com/pulumi/examples/blob/master/azure-py-aks/__main__.py

KubeDashboard is deprecated and breaks example

addon_profiles={
"KubeDashboard": {
"enabled": True,
},
},

PR #984, opened 2021-04-27
Fixup Azure-*-aks examples to exclude Addons as that's deprecated

stack72:
Fixes: #960


viveklak:
Looks good! Do you know if the azure classic ones have the add-on defined as well? Specifically - the kubedashboard add-on is deprecated.

viveklak (approved):

stack72:
They seem to be clear but I will make sure they are cleared up if not

PR #991, opened 2021-05-04
Add python SDK call example

viveklak:
Fixes pulumi/pulumi-azure-native#601


azure-py-call-azure-sdk/__main__.py
```diff
@@ -0,0 +1,52 @@
...
+from azure.core.credentials import AccessToken
+from azure.mgmt.authorization import AuthorizationManagementClient
+from pulumi_azure_native import authorization, containerregistry, resources
+
+
+class TokenCred:
+    def __init__(self, token):
+        self.token = token
+
+    def get_token(self, *scopes, **kwargs) -> 'AccessToken':
```
viveklak, line 12: This is the simplest mechanism I could find to inject the token. Very few examples and documentation on this from what I could see.

mikhailshilkov (approved):
Looks great!

PR #993, opened 2021-05-06
Aws go slackbot

ajhool:


ajhool:
@stack72 note, here: https://github.com/pulumi/examples/blob/c427abdeaf5adaf3f9ee3fbd4a70872fce06850b/aws-go-slackbot/main.go#L111

Used hardcoded subscription table name. Not sure how to convert Output to value in Go.

stack72:
Hi @ajhool 

Thanks so much for these - I took them all for a spin and they worked

With regards to the `aws-py-apigateway-lambda-serveress` the `too many requests` error goes away as soon as you hit a route. So there's something with the passthrough here that I may need to tweak

I bumped all of the pulumi versions and merged locally - these are all in the main codebase now

Thanks so much for all of this

Paul

PR #998, opened 2021-05-09
use py instead of ts in digitalocean py example

HosamAlmoghraby:
correct the ts prefix in the digitalocean-py-loadbalanced-droplets readme example page


stack72 (approved):
Thanks so much for this - sorry for the confusion here

PR #1003, opened 2021-05-13
Reproduce the condition of Pulumi hanging due to Python exception

t0yv0:


t0yv0:
Not needed any more as we have an integration test.

PR #1010, opened 2021-06-01
Fix the aws-ts-lambda-thumbnailer example

joeduffy:


joeduffy:
@stack72 It looks like there are failures in some Azure tests. Is this a known issue?

I don't think it would have been related to my changes?

stack72:
Thanks for this @joeduffy 

It's definitely not expected that the Azure examples failed here - I am going to merge this and will monitor the Azure examples afterwards

stack72 (approved):

PR #1018, opened 2021-06-11
Fix compilation of the perf test refactor

t0yv0:


t0yv0:
DigitalOcean failing as before it seems. Checking in for speed.

PR #1020, opened 2021-06-11
Integrate trace collection into the perf bench

t0yv0:


t0yv0:
This seems to be mostly working, I'm waiting for pulumi v3.4.1 release to restore links to "normal" code and then it can be checked in.

misc/test/go.mod
```diff
@@ -3,13 +3,18 @@ module github.com/pulumi/examples
...
-	github.com/pulumi/pulumi/sdk/v3 v3.0.0
+	github.com/pulumi/pulumi/sdk/v3 v3.3.1
 	github.com/stretchr/testify v1.6.1
 	gopkg.in/airbrake/gobrake.v2 v2.0.9 // indirect
 	gopkg.in/gemnasium/logrus-airbrake-hook.v2 v2.1.2 // indirect
 )
 
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v14.2.0+incompatible
+
+replace github.com/pulumi/pulumi/pkg/v3 => github.com/pulumi/pulumi/pkg/v3 v3.4.1-0.20210611165853-02c33000552c
```
stack72, line 18: why are you replacing these? Can't we just upgrade the normal deps rather than having to manage these with replacements?

misc/test/performance_test.go
```diff
@@ -3,86 +3,176 @@
...
+			maxWait := 10 * time.Minute
+			endpoint := stack.Outputs["websiteUrl"].(string)
+			assertHTTPResultWithRetry(t, endpoint, nil, maxWait, func(body string) bool {
+				return assert.Contains(t, body, "Hello, world!")
+			})
+		},
+	}
+	test := getBase(t, benchmark, opts)
 	integration.ProgramTest(t, &test)
+	computeMetricsInTest(t)
```
stack72, line 40: We can't really do this as this is the set of tests we run for ALL of our acceptance tests - 

maybe we should create a specific test for your performance test - it's a copy of what is there but named differently e,g, TestAccAwsGoS3FolderBenchmark

and we remove the build tag all in the file

misc/test/go.mod
t0yv0: Ah no good reason on using replace that's true. I was hesitating updating deps for all tests but if we can do it that's great too. For now I moved it aside so only bench tests run against pre-rel deps.

misc/test/performance_test.go
t0yv0: Hmm yeah it did run a bit aggressively. Let me try to move it to "TestMain/teardown", and note it only will run if PULUMI_TRACING_DIR is set, that's only in the perf context. It should not run in our normal acceptance tests I don't think. Let me test it in GHA.

t0yv0:
@stack72 this might be mergeable now? It's working great for me and I think I've isolated any changes to the new job only.

t0yv0:
@stack72 - sorry I'm a bit confused if you're working today or on PTO but if you are working PTAL, your concerns should be addressed here.

.github/workflows/performance_metrics_cron.yml
```diff
@@ -55,16 +60,32 @@ jobs:
...
       - uses: actions/checkout@v2
-      - name: Install Latest Stable Pulumi CLI
-        uses: pulumi/setup-pulumi@v2
+      - name: "Temporarily hack go.mod to get pre-release pulumi deps"
+        run: ./misc/test/temp_hack_go_mod.sh
+      # - name: Install Latest Stable Pulumi CLI
+      #   uses: pulumi/setup-pulumi@v2
+      - name: Install Pulumi CLI with tracing additions (temp until Latest Stable has it)
+        run: |-
+          curl -sSL https://get.pulumi.com | bash -s -- --version 3.5.0-alpha.1623852893
```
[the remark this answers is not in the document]
stack72: After speaking with anton, this is a temp measure until the next Pulumi releae

stack72 (approved):
Let's ship this and get this into the normal cron workflow so we can get something as expected

PR #1025, opened 2021-06-24
Try to run CLI metrics job on latest stable

t0yv0:


t0yv0:
This seems to work as expected now (with 3.5.1) - @stack72 just FYI, I will merge, this removes a wart.

PR #1028, opened 2021-06-26
Bug fix: Allow false to be set in the config

dorward:
The old code used the logical OR operator so that if the value in the config was false, it would be set to true. This made it impossible to set it to false in the config. This commit changes that to the nullish coalescing operator so that it will only default to true if no value is set.


leezen (approved):

leezen:
Thanks for the submission!

PR #1034, opened 2021-06-29
Add go example of kubernetes configmap rollout

mattstratton:
Adds a Go version of the kubernetes configmap rollout example

Signed-off-by: Matt Stratton <matt.stratton@gmail.com>


kubernetes-go-configmap-rollout/README.md
```diff
@@ -0,0 +1,151 @@
...
+## Running the App
+
+Follow the steps in [Pulumi Installation and Setup](https://www.pulumi.com/docs/get-started/install/) and
+[Configuring Pulumi Kubernetes](https://www.pulumi.com/docs/intro/cloud-providers/kubernetes/setup/) to
+get setup with Pulumi and Kubernetes.
+
+Install dependencies:
+
+```sh
+go get ./...
```
stack72, line 21: I am assuming you may have seen this being used in our other go examples but we should use `go mod download` here

kubernetes-go-configmap-rollout/README.md
```diff
@@ -0,0 +1,151 @@
...
+Enter a stack name: configmap-rollout-dev
+```
+
+This example will attempt to expose the `nginx` deployment to the Internet with
+a `Service` of type `LoadBalancer`. Since minikube does not support
+`LoadBalancer`, the application already knows to use type `ClusterIP` instead;
+all you need to do is to tell it whether you're deploying to minikube:
+
+```sh
+pulumi config set kubernetes-go-configmap-rollout:isMinikube <value>
```
stack72, line 37: not sure if you know but you don't need to add the project name as part of the config variable:

```
pulumi config set isMinikube <value>
```

Will auto prefix it

stack72 (approved):

PR #1035, opened 2021-06-29
Switched C# mocked unit tests to AzureNative.

pierrekhouryb:
@mikhailshilkov Updated the package reference and fixed a few related issues. Also replaced StorageAccountBelongsToResourceGroup with StorageAccountExists to work around that issue: https://github.com/pulumi/pulumi-azure-native/issues/574


mikhailshilkov (approved):
Thanks a ton for doing that @pierrekhouryb !

PR #1036, opened 2021-07-01
Use a local copy of ffmpeg

leezen:
In tests, sometimes downloading the static copy of ffmpeg errors out (possibly due to rate limiting). Switch to using a local copy of ffmpeg with the `awsx-ts-lambda-thumbnailer` example instead.


mikhailshilkov (approved):
Thank you for fixing this!

PR #1037, opened 2021-07-03
Refactor to use pytest over unittest

jandom:
@stack72 this is a re-worked unittest into pytest example – I'm not actually advocating dropping the unittest docs but I would say pytest is the more popular choice in 2021?

Also, I have further improvements for docs around code that uses 
- Unit testing programs that use Pulumi config (very commonly used, very tricky to get to work with unit tests)
- Unit testing component resources (couldn't really find any docs on that?)

### DIFF
--- a/testing-unit-py/README.md
+++ b/testing-unit-py/README.md
@@ -9,13 +9,14 @@ An example of writing mock-based unit tests with both infrastructure definition 
    ```bash
    $ python3 -m venv venv
    $ source venv/bin/activate
-   $ pip3 install -r requirements.txt
+   $ python -m pip install --upgrade pip
+   $ python -m pip install -r requirements.txt
    ```
 
 2.  Run the tests:
 
     ```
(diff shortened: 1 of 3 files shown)


jandom:
Hi there, the PR has been opened for a few days now and I'm wondering if it's helpful at all, or what to do to make it *actually* helpful rather than just helpful in my eyes? ;-)

stack72:
Sorry @jandom - I am finally catching up here! This does indeed read much nicer :) I am happy to merge this in - I need to fix a CI build which runs the pip3 install command but it will be ok

Tests are running now

jandom:
Awesome stuff @stack72  – I wasn't sure what was happening there. It's a pretty big shift to recommend pytest over unittest but I think that's the preferred one. 

I was also thinking of writing up how to test composite components and stacks using pulumi.config? These are both useful advanced test cases that are not super-straightforward and any serious user will bump into them

PR #1041, opened 2021-07-08
Benchmark for scaling in N of independent resources

t0yv0:
Fixes https://github.com/pulumi/home/issues/1467

Here are some numbers from the CI run:

- https://metabase.corp.pulumi.com/question/764
- https://metabase.corp.pulumi.com/question/767

The findings that stand out:

- at worst linear scaling observed
 
- for most slow numbers, the time is dominated by `time_pulumi_api_ms` - strongly indicating that we're simply being too chatty here, perhaps the `deadweight` diff handling and saving to service state is dominating this

- preview-initial is 38s with very small `time_pulumi_api_ms` - why? Hypothesis: installing networked dependencies and compiling, limited visibility into this

- destroy is massively slow, why would that be

In brainstorming with @lukehoban earlier we wanted to focus on `pulumi-update-empty` numbers for the definitive benchmark with alert thresholds.

Does this PR sounds like a reasonable benchmark? Should we remove deadweight and increase resource counts further? Or at least experiment with that?

# How to run these benchmarks to update data

1. Kickoff https://github.com/pulumi/examples/actions/workflows/performance_metrics_cron.yml from the right branch; this puts data in S3

2. Manually invoke a [lambda](https://app.pulumi.com/pulumi/cli-performance-metrics/cli-performance-metrics-prod/resources/latest/urn:pulumi:cli-performance-metrics-prod::cli-performance-metrics::pulumi:service:CliPerformanceDataLoader$aws:lambda/function:Function::cliperfloader-lambda) function

3. Alternatively wait 2h it will sync


stack72 (approved):

cs-many-resources/.gitignore
```diff
@@ -0,0 +1,353 @@
+## Ignore Visual Studio temporary files, build results, and
```
lukehoban, line 1: Should this be in a root folder?  I don't think it has been standard to include this `.gitignore` elsewhere?

cs-many-resources/Dummy.cs
```diff
@@ -0,0 +1,15 @@
+using Pulumi;
```
lukehoban, line 1: I'm a bit unsure about this being in the `examples` repo - that repo is a user-facing set of examples, and this is not a real example. 

If we do decide we really need this to live in `examples`, I'd suggest nesting it inside some "infrastructure" or "misc" like folder so it's not as user-facing.

cs-many-resources/MyStack.cs
```diff
@@ -0,0 +1,40 @@
...
+    [Output]
+    public Output<int> ResourceCount { get; set; }
+
+    [Output]
+    public Output<int> ResourcePayloadBytes { get; set; }
+
+    public MyStack()
+    {
+        int resourceCount = GetEnv("RESOURCE_COUNT", 64);
+        int resourcePayloadBytes = GetEnv("RESOURCE_PAYLOAD_BYTES", 1024);
```
lukehoban, line 19: Nit: Any reason to use env vars instead of Pulumi config?

cs-many-resources/Dummy.cs
t0yv0: @stack72 thoughts on where to find a home for this?

1. new `pulumi-benchmarks` repo
2. put into `pulumi/pulumi`

Moving the code is trivial, but I'm a little unsure about cloning the YAML Github Action definitions that run the code, I think you mentioned Paul that those are really managed from some central code-generating repository, was it https://github.com/pulumi/ci-mgmt ? Could you give me a few pointers so I get it done the way we like it.

t0yv0: For the immediate need right now I can definitely nest this so as not to distract the users looking for actual examples.

cs-many-resources/MyStack.cs
t0yv0: Good idea, fixed.

pgavlin:
Note that benchmarking using component resources will not give the same results as benchmarking using custom resources. A single component resource requires 2 calls to the engine--one call to `RegisterResource` and one to `RegisterResourceOutputs`--where a single custom resource only requires the call to `RegisterResource`. It may be worth capturing this difference in another benchmark.

t0yv0:
@pgavlin that's great - I think I'd like to check this in as-is, but I'm stuffing these ideas into:

- https://github.com/pulumi/home/issues/1500
- https://github.com/pulumi/home/issues/1501

And more generally https://github.com/pulumi/home/issues/1499

I'll specifically need your input in one of these, I will tag you.

PR #1044, opened 2021-07-15
Update TS Step Functions to use Callback Functions Instead of Serverless Functions

phillipedwards:
This is the update the example to use the aws.lambda.callbackFunction resource instead of the deprecated aws.serverless.Function.


aws-ts-stepfunctions/index.ts
```diff
@@ -6,63 +6,63 @@ import * as pulumi from "@pulumi/pulumi";
...
 });
 
-const helloFunction = new aws.serverless.Function(
-    "helloFunction",
-    { role: lambdaRole },
-    (event, context, callback) => {
-        callback(null, "Hello");
-    },
-);
+const helloFunction = new aws.lambda.CallbackFunction(
```
pierskarsenbarg, line 46: You can just use `aws.lambda.Function` instead here. I think the `CallbackFunction` is for this situation https://www.pulumi.com/docs/guides/crosswalk/aws/lambda/#specifying-attributes-on-your-magic-lambda-function

phillipedwards: Does ```aws.lambda.Function``` support in-line and/or magic functions in node? My understanding was it supported a type of deployment package, whether that be a local app, s3 bucket, etc.

pierskarsenbarg: Got it. Yes you're right

PR #1048, opened 2021-07-19
Switched arguments names from CamelCase to snake_case

jancespivo:
Naming conventions should be consistent (at least in a file) and snake_case is more "pythonic" :)


stack72 (approved):
Thanks so much for this!

PR #1053, opened 2021-07-24
Fix minimum pulumi-tls version in python requirements

eliaskousk:
For some examples the minimum `pulumi-tls` package version currently set is `v4.2.0` which doesn't exist.
The latest version that exists is `v4.0.0`.

I updated the files using the `update_requirements.txt` script which had already set and thus also bumped the minimum `pulumi` package version to `v3.5.1` in ` misc/benchmarks/py-many-resources`. Let me know if you want this in a separate commit.

Signed-off-by: Elias Kouskoumvekakis <eliaskousk@gmail.com>


stack72:
Thanks so much for this - I'm sorry this caused an issue :/

eliaskousk:
Thank you a lot Paul for the quick merge. And for your great work on Pulumi, it has helped greatly on some projects.

stack72:
Thanks @eliaskousk - very kind of you to say!

PR #1055, opened 2021-07-29
Include aws py provisioners test

t0yv0:


aws-py-ec2-provisioners/__main__.py
```diff
@@ -12,12 +13,22 @@
...
+
+    try:
+        key = base64.b64decode(key.encode('ascii')).decode('ascii')
+    except:
+        pass
+
     if key.startswith('-----BEGIN RSA PRIVATE KEY-----'):
         return key
+
     return key.encode('ascii')
```
t0yv0, line 29: I was confused by along time here, eventually filed https://github.com/pulumi/pulumi/issues/7691 - we do not seem to support `bytes` object, and we were missing b64 decoding step. Integration test passes a value that is a secret string but is actually b64-encoded.

aws-py-ec2-provisioners/provisioners.py
```diff
@@ -65,6 +65,9 @@ def connect(conn: ConnectionArgs) -> paramiko.SSHClient:
 class ProvisionerProvider(dynamic.ResourceProvider):
     __metaclass__ = abc.ABCMeta
 
+    def ignore_properties(self) -> List[str]:
```
t0yv0, line 68: Is there a more idiomatic fix? The integration test was verifying that the empty preview remains empty, it wasn't true because the `results` property appeared to be being removed for the diff on `cat_config = provisioners.RemoteExec('cat-config',...)`

misc/test/aws_test.go
```diff
@@ -346,6 +354,9 @@ func TestAccAwsTsEc2Provisioners(t *testing.T) {
 		KeyName: aws.String(keyName),
 	})
 	assert.NoError(t, err)
+	if err != nil {
```
t0yv0, line 357: This is just so when running locally it fails faster if AWS creds are borked.

aws-py-ec2-provisioners/provisioners.py
```diff
@@ -157,7 +166,7 @@ def on_create(self, inputs: Any) -> Any:
 # RemoteExec runs remote one or more commands over an SSH connection. It returns the resulting
 # stdout and stderr from the commands in the results property.
 class RemoteExec(dynamic.Resource):
-    results: pulumi.Output[list]
+    results: pulumi.Output[List[Any]]
```
t0yv0, line 169: Temporary fix - next pulumi (after 3.9.1)  will accept the original program.

aws-py-ec2-provisioners/provisioners.py
justinvp: How does the TS version compare?

justinvp (approved):

t0yv0:
Looks like I accidentally merged this. Following up shortly, looks like the TS code does not include results in the subject of the diff already so it avoids this situation, I'll update Python to match.

t0yv0:
OK never mind, leaving as is as hitting diminishing returns.

TS version uses a different structure. It uses this:

```
// Provisioner lets a custom action run the first time a resource has been created. It takes as input
// a dependent property. Anytime its value changes, the resource is replaced and will re-run its logic.
export class Provisioner<T, U> extends pulumi.dynamic.Resource {
```

With T=ConnectionArgs. So it only runs diff on ConnectionArgs.

In Python, the structure of resources is different. I tried `ignore_changes` option but it does not work. The DynamicProvider cannot reference ResourceOptions and therefore runs diff over everything, and accidentally runs diff over results. Therefore extending the provider in the examples with its own ignore_changes analogue seems reasonable.

PR #1059, opened 2021-08-12
Include missing dependencies for testing-unit-ts

leezen:


leezen:
Interesting -- this somehow still fails the test and it claims it cannot find the `typescript` module...

leezen:
Ah, I think it's b/c the test is using the global `mocha` and `ts-node` -- I don't think we actually need to do that, so I'll update the workflow.

mikhailshilkov (approved):
LGTM, thank you for the cleanup

PR #1063, opened 2021-08-20
Example: how to unit test a component resource

jandom:
An example that would have really helped me in the past – when I started breaking things up into Component Resources, I really missed any notes on how to test. It's possible to figure it out but some "best practice" would be great

https://www.pulumi.com/docs/intro/concepts/resources/#components

Especially in the tutorial on Component Resources, there is no mention on unit testing them in isolation
https://www.pulumi.com/docs/tutorials/aws/s3-folder-component/

I may have done too much here – sub-foldered into `mocha/` , changed how mocha is called (to use the node_modules, rather than a global install)


testing-unit-ts/mocha/bucket_pair.ts
```diff
@@ -0,0 +1,23 @@
...
+        this.contentBucket = new aws.s3.Bucket("contentBucket", {
+            bucket: contentBucketName,
+        }, { parent: this });
+
+        this.logsBucket = new aws.s3.Bucket("logsBucket", {
+            bucket: logsBucketName,
+        }, { parent: this });
+
+        // Register output properties for this component
+        this.registerOutputs();
```
pierskarsenbarg, line 21: For completeness, you should register the outputs like this:

```suggestion
        this.registerOutputs({
            contentBucket: this.contentBucket,
            logsBucket: this.logsBucket
        });
```

This is so Pulumi knows that everything has been done (I'll find the more technical term later on)

pierskarsenbarg (changes requested):
Looks good, but one comment/suggestion for completeness

testing-unit-ts/mocha/bucket_pair.ts
jandom: (Will implement this change obviously, it's super-easy – big thanks for your review ;-)!)

pierskarsenbarg:
There's more information in [this link](https://www.pulumi.com/docs/intro/concepts/resources/#registering-component-outputs), but the important bit is as follows:

```
The call to registerOutputs also tells Pulumi that the resource is done registering children and should be considered fully constructed, so—although it’s not enforced—the best practice is to call it in all components even if no outputs need to be registered.
```

jandom:
Nice, many thanks for the merge @pierskarsenbarg

pierskarsenbarg:
Thank you for the PR 😀

PR #1066, opened 2021-08-31
Use Origin Access Identity and lock down S3 permissions on the AWS static website example

lita:
I used this example to launch our single-page app, but felt like giving public read access wasn't awesome from a security standpoint. We ended up creating an Origin Access Identity and creating a policy to only allowing Cloudfront to access the S3 bucket. Thought I share the solution as a PR as it seems like a better security posture for a canonical example.


jandom:
This is a really nice improvement – I never remember about the "little details" like not leaving S3 buckets public :P

jandom:
(ship it!)

jkodroff (approved):

jkodroff:
@lita Thanks so much for the contribution and my apologies for the slow response!

PR #1067, opened 2021-09-02
Add Python kubernetes exposed deployment example

katcosgrove:
Adding a Python version of the kubernetes exposed deployment example


kubernetes-py-exposed-deployment/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pulumi==3.11.0
```
stack72, line 1: we usually use ranges than go between major versions so that we don't need to keep bumping this :)

stack72 (approved):
left 1 nit that you can push outside of this PR :D Thank you!

PR #1077, opened 2021-09-17
Add a temporary cron job with output values enabled

justinvp:
The output values feature is currently disabled by default. This duplicates the existing `providers` job with the feature enabled via `PULUMI_ENABLE_OUTPUT_VALUES` envvar.

Once we enable the feature by default, we can remove this job here.


emiliza (approved):

justinvp:
> should go be in the test matrix as well?

@EvanBoyle, no, this job was only temporary and I just opened #1099 to remove it.

PR #1084, opened 2021-09-29
Attempt to reduce test parallelism for AWS Python tests

leezen:
The last few test runs fail when setting up the dependencies with an EOF error, which I suspect may be due to disk space. Attempt to reduce the parallelism to see if that addresses the issue.


stack72 (approved):
lets give this a shot and see

PR #1095, opened 2021-10-07
Update Python examples to use fn_output forms where appropriate

t0yv0:
Context: pulumi/pulumi#7948

Updates Python examples to use $fn_output feature.

### DIFF
--- a/aws-py-django-voting-app/__main__.py
+++ b/aws-py-django-voting-app/__main__.py
@@ -228,9 +228,8 @@ def get_registry_info(creds):
         username=username,
         password=password)
 
-app_registry = app_ecr_repo.registry_id \
-                           .apply(aws.ecr.get_credentials_output) \
-                           .apply(get_registry_info)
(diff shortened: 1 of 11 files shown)


t0yv0:
Completed eye-balling all Python examples ```Py   67/67``` - since human error is possible, could have omitted something.

This will need a fresh look. Also I am thinking that we should fix pulumi/pulumi#8172 as a pre-requisite and fix the code to use direct calls and no workarounds. In a lot of the places 8172 is a show-stopper.

t0yv0:
pulumi/pulumi#8172 fixed. Previous CI failures were hitting it. Let's see it if passes this time.

t0yv0:
CI failure indicating pulumi/pulumi#8172 has not rolled out to pulumi-azure-native yet.

t0yv0:
The fix is rolled out and we have a ✅  PTAL.

justinvp (approved):

PR #1096, opened 2021-10-08
Update Node JS examples to use the new $fnOutput feature

t0yv0:
Context: pulumi/pulumi#7948

Updates Node JS examples to use $fnOutput feature. 

This PR will become merge-able when https://github.com/pulumi/pulumi/pull/8047 propagates into providers.

Progress counters:

```
Node 151/151
```

### DIFF
--- a/aws-ts-appsync/iam.ts
+++ b/aws-ts-appsync/iam.ts
@@ -5,7 +5,7 @@ import * as pulumi from "@pulumi/pulumi";
 
 export function createIamRole(name: string, table: aws.dynamodb.Table) {
     const role = new aws.iam.Role(`${name}-role`, {
-        assumeRolePolicy: aws.iam.getPolicyDocument({
+        assumeRolePolicy: aws.iam.getPolicyDocumentOutput({
                 statements: [{
                     actions: ["sts:AssumeRole"],
                     principals: [{
@@ -14,17 +14,17 @@ export function createIamRole(name: string, table: aws.dynamodb.Table) {
                     }],
                     effect: "Allow",
                 }],
-            }, { async: true }).then(doc => doc.json),
+            }, { async: true }).json,
     });
 
     const policy = new aws.iam.Policy(`${name}-policy`, {
-        policy: table.arn.apply(arn => aws.iam.getPolicyDocument({
+        policy: aws.iam.getPolicyDocumentOutput({
             statements: [{
                 actions: ["dynamodb:PutItem", "dynamodb:GetItem"],
-                resources: [arn],
+                resources: [table.arn],
                 effect: "Allow",
             }],
-        }, { async: true }).then(doc => doc.json)),
+        }, { async: true }).json,
     });
 
     const attachment = new aws.iam.RolePolicyAttachment(`${name}-rpa`, {
--- a/aws-ts-appsync/iam.ts
+++ b/aws-ts-appsync/iam.ts
@@ -14,7 +14,7 @@ export function createIamRole(name: string, table: aws.dynamodb.Table) {
                     }],
                     effect: "Allow",
                 }],
-            }, { async: true }).json,
+            }).json,
     });
 
     const policy = new aws.iam.Policy(`${name}-policy`, {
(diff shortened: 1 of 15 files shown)


t0yv0:
Continuing the sweep ```Node 60/151``` - TBC.

t0yv0:
Finished ```Node 151/151```, looking at int and test failures.

aws-ts-netlify-cms-and-oauth/cms/infrastructure/index.ts
```diff
@@ -98,7 +98,7 @@ if (config.certificateArn === undefined) {
     }, { provider: eastRegion });
 
     const domainParts = getDomainAndSubdomain(config.targetDomain);
-    const hostedZoneId = aws.route53.getZone({ name: domainParts.parentDomain }, { async: true }).then(zone => zone.zoneId);
+    const hostedZoneId = aws.route53.getZoneOutput({ name: domainParts.parentDomain }, { async: true }).zoneId;
```
pgavlin, line 101: These feel a bit funny, tbh, as they're trading off less idiomatic JS/TS in favor of the lifted property accessors on `Output`s

t0yv0: Is there a problem with lifted property accessors, are they about to go away? What is idiomatic Pulumi? 

I forget to use them typically but I get review comments suggesting to use them:

https://github.com/pulumi/pulumi-hugo/pull/780#discussion_r735620716

t0yv0:
Looks like the failure is about jumping the gun on DigitalOcean provider. ```index.ts(31,48): error TS2339: Property 'getKubernetesClusterOutput' does not exist on type 'typeof import("/tmp/p-it-fv-az216-1-digitaloce-8cbf8301-252923268/node_modules/@pulumi/digitalocean/index")'.```

aws-ts-netlify-cms-and-oauth/cms/infrastructure/index.ts
pgavlin: > What is idiomatic Pulumi?

This neatly summarizes my uneasiness with output functions in general. I don't see the need to use them unless I have an input that is itself an output--which is not the case in this example--as the "normal" versions are typically more idiomatic in the target language. In the past, we've tried to be as idiomatic as possible in the target language so that example code will look as familiar as possible to the end user.

As far as when to use lifted property accessors, I would say that any time the thing you're working with is already an output or is "naturally" an output (e.g. it is a resource output or the result of a call to an output function), then lifted property accessors are the way to go. If what you're working with is not naturally an output, it feels funny to force it to be an output just for the sake of lifted property accessors.

t0yv0: Gotcha so a program that does not use outputs is preferable to one that does. I can revert.

t0yv0: Reverted all similar changes.

t0yv0:
Removed digitalocean change as it's not yet updated (nor planning to update). Merge master to fix TestAccAwsJsSqsSlack. TestAccAwsPyResources believed flaky, trying again.

t0yv0:
PTAL, CI finally passing after all the fixes, and feedback got addressed

aws-ts-appsync/iam.ts
```diff
@@ -14,17 +14,17 @@ export function createIamRole(name: string, table: aws.dynamodb.Table) {
                     }],
                     effect: "Allow",
                 }],
-            }, { async: true }).then(doc => doc.json),
+            }, { async: true }).json,
```
justinvp, line 17: We don't need the `async: true` invoke option, right? (Applies throughout)

```suggestion
            }).json,
```

justinvp (approved):

aws-ts-appsync/iam.ts
t0yv0: Love this comment, thank you! Turns out

```
Invoking data source functions
     * synchronously is deprecated.  The ability to do this will be removed at a later point in
     * time.
```

Definitely need to remove.

pgavlin (approved):

t0yv0:
https://github.com/pulumi/examples/issues/1114 flaky test

PR #1106, opened 2021-10-31
Fixes/azure-ts-aks-helm

PaulusTM:
The Azure AKS Helm example does not work anymore. 
This PR fixes the demo.


stack72 (approved):

stack72:
Thanks so much for this!!

PR #1111, opened 2021-11-04
Try upgrading to Node 14.x

t0yv0:
I did not find docs about which version of node we intentionally test on. Trying to upgrade to 14.x, motivation is that I found this failure on CI:

```
Failing  TestAccAwsJsSqsSlack

   error p-retry@5.0.0: The engine "node" is incompatible with this module. Expected version "^12.20.0 || ^14.13.1 || >=16.0.0". Got "13.14.0"
   error Found incompatible module.
```

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -151,7 +151,7 @@ jobs:
         go-version:
         - 1.16.x
         node-version:
-        - 13.x
+        - 14.x
         platform:
         - ubuntu-latest
         python-version:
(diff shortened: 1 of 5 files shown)


t0yv0:
Alternatively I can cancel this test until we can try to find a working dependency set for it.

Frassle:
Given 13 is out of support (https://nodejs.org/en/about/releases/) we should probably upgrade to at least 14, or even 16 which is still in active support.

t0yv0:
@Frassle if you give me a ✔️ I can merge in. This only affects our CI tests afaik, and can easily be reverted if we need to, say, test on 12x.

Frassle (approved):

PR #1116, opened 2021-11-09
Fix error in ProvisionerProvider.diff() when inputs are pulumi.output.Unknown

jamesfmilne:
This patch adds a helper function to handle pulumi.output.Unknown objects when performing a diff on a ProvisionerProvider-based resource.

This in turn fixes issues with `pulumi preview` and `pulumi up` when using these providers in a stack when any of the inputs changes.


mjeffryes:
I'm sorry we let this sit so long, but I think this example has changed significantly and the code no longer merges so I'm going to close for now. Feel free to reopen with an updated change (we're now watching this repository for contributions more regularly)

PR #1122, opened 2021-11-23
Add Crosswalk API Gateway multi-language examples

danielrbradley:
Create updated examples for AWS API Gateway to go along with updating the main guide (https://github.com/pulumi/pulumi-hugo/pull/867).

- Create comparible examlpes in Typescript, Python and Go
- Create static (S3), Lambda and HTTP proxy routes
- Create Cognito and Lambda authorizers
- All examples and instructions tested locally


aws-apigateway-go-routes/handler/handler.go
```diff
@@ -0,0 +1,17 @@
+package main
```
mikhailshilkov, line 1: We probably want to add copyright to all code files as in https://github.com/pulumi/examples/blob/master/azure-go-aks/main.go#L1

aws-apigateway-py-routes/__main__.py
```diff
@@ -0,0 +1,47 @@
+import pulumi
+import pulumi_aws as aws
+import pulumi_aws_apigateway as apigateway
+import lambdas_
```
mikhailshilkov, line 4: Is this trailing underscore required in the file name?

aws-apigateway-go-routes/main.go
```diff
@@ -0,0 +1,146 @@
...
+			Runtime: lambda.RuntimeGo1dx,
+			Code: pulumi.NewAssetArchive(map[string]interface{}{
+				".": pulumi.NewFileArchive("./bin/handler.zip"),
+			}),
+			Handler: pulumi.String("handler"),
+			Role:    role.Arn,
+		}, pulumi.DependsOn([]pulumi.Resource{logPolicy}))
+		if err != nil {
+			return err
+		}
```
mikhailshilkov, line 71: Would it make sense to move lambdas to a separate file/function similar to Python?

mikhailshilkov (commented):
These look great to me. Is C# coming?

aws-apigateway-py-routes/__main__.py
danielrbradley: Seemed like it broke without it - think lambda is a reserved keyword in python.

mikhailshilkov: But not `lambdas` I guess?

danielrbradley: Yup - the suffix can be removed fine 👍

aws-apigateway-go-routes/main.go
```diff
@@ -0,0 +1,206 @@
...
+		apiId := restAPI.Api.ApplyT(func(api *awsapigateway.RestApi) pulumi.StringOutput {
+			return api.ID().ToStringOutput()
+		}).ApplyT(func(id interface{}) string {
+			return id.(string)
+		}).(pulumi.StringOutput)
+		stageName := restAPI.Stage.ApplyT(func(stage *awsapigateway.Stage) pulumi.StringOutput {
+			return stage.StageName
+		}).ApplyT(func(stageName interface{}) string {
+			return stageName.(string)
+		}).(pulumi.StringOutput)
```
mikhailshilkov, line 148: This double-apply looks a bit wild. Aren't we getting the stageName output from the first ApplyT? Same question to the `apiId` above.

mikhailshilkov (approved):
LGTM with one question

aws-apigateway-go-routes/main.go
danielrbradley: Yes it does appear to be required. The first apply retrieve the sub-property, which itself is an output needing to be resolved. The second apply is to allow us to convert the unknown value to the correct type.

lukehoban:
🎉

Issue #1123, opened 2021-11-24
Can't export kubeconfig when following the guide azure-cs-aks


d1mnewz:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details
Trying to follow https://www.pulumi.com/registry/packages/azure-native/how-to-guides/azure-cs-aks/
https://github.com/pulumi/registry/issues/409

![image](https://user-images.githubusercontent.com/11834902/143279136-c99e3f76-6f05-43d2-8458-6e1073894278.png)

### Steps to reproduce

Issue when following this guide:

> dimko@DESKTOP-KVKFCQP:~/.pulumi/test/azure-cs-aks$ pulumi stack output kubeconfig --show-secrets > kubeconfig.yaml
> error: current stack does not have output property 'kubeconfig'

It's gotta be 
> pulumi stack output KubeConfig --show-secrets > kubeconfig.yaml

since property name is case sensetive. 

![image](https://user-images.githubusercontent.com/11834902/143042900-3eaf1234-647d-49c6-b797-6cd1b73eda57.png)

Expected: the command from doc works 
Actual: the command does not :)

leezen:
Thanks for submitting both the bug and the corresponding PR

leezen:
Ah, looking at your PR, it seems like there's already https://github.com/pulumi/examples/pull/1127 which does the opposite (aligns the output to the README), so I'll go ahead and close this as fixed and close the PR as well.

PR #1124, opened 2021-11-24
Change kubeconfig to be case-sensetive.

d1mnewz:
Should be resolving https://github.com/pulumi/examples/issues/1123


leezen:
Superseded by https://github.com/pulumi/examples/pull/1127

PR #1128, opened 2021-11-27
Improvements to `testing-unit-ts`

henriiik:
Hello!

Whilst trying to learn testing with pulumi I used the `testing-unit-ts` as a base via `pulumi new`. Whilst trying to base a project on that I realised that there were some improvement I could make to make it easier for the next person. This MR is split 3 commits to make it easier to review each change in isolation.

1. 881a42022f2dc26c56aa4f476ac9089e108ecfd0 add npm test script to testing-unit-ts
    - Added a script to run all the tests in the project, as you would want in a ci pipeline
    - Renamed the test files so they consistently end with `.tests.ts` to make it easy to write a glob pattern.
2. 17bca955b1c0aa65cf1deac9f6e2fe75a72fbf87 share mocks
    - Extracted the mocks into a separate shared file.
    - When running multiple test files each call to `pulumi.runtime.setMocks` (which happens are module load time) will overwrite what ever was previously set which can lead to very confusing behavior for someone who is not aware of that.
3. 7ba8c1ae7c0730120195836509a17c2cb0aac257 convert to tests to async functions
    - Convert tests to use async functions
        - With a small `unwrap` helper, it is possible to write tests as async functions instead of as regular function with a `done` callback. This has some advantages:
            - assert works as expected, if an assertion fails mocha catches it rather than having to use try/catch to call `done` with an error.
            -  fewer nested functions makes code easier to read and write.

If you are interested there are more changes that could be made to make this more appropriate as a base for new projects ex. extract mocha config into a `.mocharc` file (instead of having it inline in the call to `mocha -r ts-node/register`)


mjeffryes:
I'm sorry we let this sit so long, but I think this example has changed significantly and the code no longer merges so I'm going to close for now. Feel free to reopen with an updated change (we're now watching this repository for contributions more regularly)

PR #1149, opened 2022-01-12
Add API Gateway-EventBridge examples

cnunciato:
Adds examples in TypeScript and Python of integrating API Gateway with EventBridge and Lambda. (Intended to accompany https://github.com/pulumi/pulumi-hugo/pull/955.)


aws-py-apigatewayv2-eventbridge/README.md
```diff
@@ -0,0 +1,71 @@
+# API Gateway to EventBridge
+
+[![Deploy with Pulumi](https://get.pulumi.com/new/button.svg)](https://app.pulumi.com/new?template=https://github.com/pulumi/examples/blob/master/aws-apigateway-go-routes/README.md)
```
cnunciato, line 3: Erp, copy-pasta here and in the other example -- fixing.

aws-py-apigatewayv2-eventbridge/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+    ```bash
+    pulumi config set aws:region us-west-2
+    ```
+
+1. Install Python dependencies and run Pulumi:
+
+    ```bash
+    python3 -m venv venv
+    venv/bin/pip install -r requirements.txt
```
nimbinatus, line 33: It's better to activate the venv before running anything. Also, note that pip will automatically be pip3 when you use python3 to create the venv; there's also other flags to ensure you're using Python 3 if you're working on a different setup than the base one, but I doubt someone who has that all set up will be going by these directions verbatim 😉  
```suggestion
    python3 -m venv venv
    source venv/bin/activate
    pip install -r requirements.txt
```

aws-py-apigatewayv2-eventbridge/__main__.py
```diff
@@ -0,0 +1,132 @@
...
+api = aws.apigatewayv2.Api("example",
+    protocol_type="HTTP"
+)
+
+# Create a stage and set it to deploy automatically.
+stage = aws.apigatewayv2.Stage("stage",
+    api_id=api.id,
+    name=pulumi.get_stack(),
+    auto_deploy=True
+);
```
nimbinatus, line 17:
```suggestion
)
```

aws-py-apigatewayv2-eventbridge/__main__.py
```diff
@@ -0,0 +1,132 @@
...
+                "Principal": {
+                    "Service": "apigateway.amazonaws.com",
+                },
+            },
+        ],
+    }),
+    managed_policy_arns=[
+        "arn:aws:iam::aws:policy/AmazonEventBridgeFullAccess",
+    ],
+);
```
nimbinatus, line 45:
```suggestion
)
```

aws-py-apigatewayv2-eventbridge/__main__.py
```diff
@@ -0,0 +1,132 @@
+# Copyright 2016-2021, Pulumi Corporation.  All rights reserved.
```
nimbinatus, line 1: There's some [PEP8](https://www.python.org/dev/peps/pep-0008/) whitespace stuff (blank line counts, indentation) that I'd change here. They're not crucial/blockers as the code certainly will still run, but I'd recommend following them if we can as some Python devs might give this a bit of side-eye. Would you like me to fix them?

aws-py-apigatewayv2-eventbridge/__main__.py
```diff
@@ -0,0 +1,132 @@
...
+    # event source matches pattern defined on the EventRule, and the
+    # Detail expression, which just forwards the body of the original
+    # API Gateway request (i.e., the uploaded document).
+    request_parameters={
+        "EventBusName": bus.name.apply(lambda name: name),
+        "Source": "my-event-source",
+        "DetailType": "my-detail-type",
+        "Detail": "$request.body",
+    },
+);
```
nimbinatus, line 66:
```suggestion
)
```

aws-py-apigatewayv2-eventbridge/__main__.py
```diff
@@ -0,0 +1,132 @@
...
+
+# Finally, define the route.
+route = aws.apigatewayv2.Route("route",
+    api_id=api.id,
+    route_key="POST /uploads",
+    target=integration.id.apply(lambda id: f"integrations/{id}"),
+);
```
nimbinatus, line 73:
```suggestion
)
```

aws-py-apigatewayv2-eventbridge/__main__.py
```diff
@@ -0,0 +1,132 @@
...
+        ".": pulumi.FileArchive('./api')
+    })
+)
+
+# Create an EventBridge target associating the event rule with the function.
+lambda_target = aws.cloudwatch.EventTarget("lambda-target",
+    arn=lambda_function.arn,
+    rule=rule.name,
+    event_bus_name=bus.name,
+);
```
nimbinatus, line 121:
```suggestion
)
```

aws-py-apigatewayv2-eventbridge/__main__.py
```diff
@@ -0,0 +1,132 @@
...
+
+# Give EventBridge permission to invoke the function.
+lambda_permission = aws.lambda_.Permission("lambda-permission",
+    action="lambda:InvokeFunction",
+    principal="events.amazonaws.com",
+    function=lambda_function.arn,
+    source_arn=rule.arn,
+);
```
nimbinatus, line 129:
```suggestion
)
```

aws-py-apigatewayv2-eventbridge/__main__.py
```diff
@@ -0,0 +1,132 @@
...
+
+# Export the API Gateway URL to give us something to POST to.
+pulumi.export("url", pulumi.Output.concat(api.api_endpoint, "/", stage.name))
```
nimbinatus, line 132:
```suggestion
pulumi.export("url", pulumi.Output.concat(api.api_endpoint, "/", stage.name))

```

aws-py-apigatewayv2-eventbridge/__main__.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # The body of the request to be sent to EventBridge. Note the
+    # event source matches pattern defined on the EventRule, and the
+    # Detail expression, which just forwards the body of the original
+    # API Gateway request (i.e., the uploaded document).
+    request_parameters={
+        "EventBusName": bus.name.apply(lambda name: name),
+        "Source": "my-event-source",
+        "DetailType": "my-detail-type",
+        "Detail": "$request.body",
```
nimbinatus, line 64: Should this be `$request.body.name` based on https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-parameter-mapping.html#http-api-mapping-request-parameters ?

(Also, I _am_ reading that right that you don't need curly braces since it's only one variable, right? It's kinda itchy in Python 😆  )

aws-py-apigatewayv2-eventbridge/api/handlers.py
```diff
@@ -0,0 +1,6 @@
+import json
+
+def capture_order(event, context):
+    # For now, just log the event, including the uploaded document.
+    # That'll be enough to verify everything's working.
+    print(json.dumps({ "source": event["source"], "detail": event["detail"] }))
```
nimbinatus, line 6:
```suggestion
    print(json.dumps({"source": event["source"], "detail": event["detail"]}))

```

aws-ts-apigatewayv2-eventbridge/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    // The body of the request to be sent to EventBridge. Note the
+    // event source matches pattern defined on the EventRule, and the
+    // Detail expression, which just forwards the body of the original
+    // API Gateway request (i.e., the uploaded document).
+    requestParameters: {
+        EventBusName: bus.name,
+        Source: "my-event-source",
+        DetailType: "my-detail-type",
+        Detail: "$request.body",
```
nimbinatus, line 69: Same question as with the Python one for `$request.body.name`

nimbinatus (changes requested):

aws-py-apigatewayv2-eventbridge/__main__.py
cnunciato: Gah -- habit! 🙏

aws-py-apigatewayv2-eventbridge/api/handlers.py
cnunciato: We should totally add this stuff to `make lint`.

aws-py-apigatewayv2-eventbridge/__main__.py
cnunciato: > Should this be $request.body.name

I'm passing the whole request body through in this case, so I think this is okay.

 > you don't need curly braces

I don't think I follow. Which braces do you mean? The ones around these four properties? If I take those out, I get this:

image

... but if there's something weird-looking or otherwise non-idiomatic here, definitely let me know and I'm happy to change.

aws-ts-apigatewayv2-eventbridge/index.ts
cnunciato: Yep, I think this is okay.

cnunciato:
@nimbinatus Thanks for reviewing these! I just added a V1 example as well -- it's TS-only, but if you wouldn't mind giving it the once-over, I'd appreciate it. I've verified it works as well and checked against the README.

aws-py-apigatewayv2-eventbridge/__main__.py
cnunciato: Oh sorry, yes -- definitely feel free to adjust anything you think needs adjusting space-wise. Thank you!

aws-ts-apigateway-eventbridge/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+1. Verify the deployment with `curl`:
+
+    With invalid POST data:
+
+    ```bash
+    curl --data '{"some-invalid-property-name": "Chris"}' --header "Content-Type: application/json" "$(pulumi stack output url)/uploads"
+
+    HTTP/2 400
+    {"message": "Invalid request body"
```
nimbinatus, line 50: I'm guessing this is the full message so it should close?
```suggestion
    {"message": "Invalid request body"}
```

cnunciato: Ah yes, bad paste on my part. Thanks!

aws-py-apigatewayv2-eventbridge/__main__.py
cnunciato: Ah, I see. Yes:

> I think, if I'm reading everything right, that string gets passed as is to AWS, which then does substitution, rather than Python doing the substitution.

That's correct. This is just a raw string that's passed to AWS, and in this case, it just means "the full request body". You can see it referenced this way under Mapping Request Parameters:

https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-aws-services.html

image

nimbinatus (approved):
Approving to unblock; only thing was that copy-paste curly brace, so I'd say you're good to go.

aws-ts-apigateway-eventbridge/README.md
```diff
@@ -0,0 +1,81 @@
...
```
cnunciato, line 50: Re-appylying this suggestion -- I must've misclicked. 🙄 

```suggestion
    {"message": "Invalid request body"}
```

PR #1160, opened 2022-01-31
Updated policy example to use azure native

pierskarsenbarg:


mjeffryes:
Closing as stale since this has not been touched in a few years and no longer merges cleanly.

Issue #1119, opened 2021-11-15
TestAccAzureGoCallAzureSdk: Reenable test


justinvp:
TestAccAzureGoCallAzureSdk is failing with the following and has been temporarily disabled:

```
  azure-native:authorization:RoleAssignment (access-from-cluster):
219
    error: autorest/azure: Service returned an error. Status=400 Code="UnmatchedPrincipalType" Message="The PrincipalId 'e7d1b8ea596b44be890f8355a351244c' has type 'ServicePrincipal' , which is different from specified PrinciaplType 'User'."
```

This issue tracks fixing the test and reenabling it.

PR #1167, opened 2022-02-11
Fix azure-go-call-azure-sdk test

jkisk:
Remove outdated comment and update test to use Service Principal.

Fixes: #1119


justinvp (approved):
LGTM. Thanks!

PR #1173, opened 2022-02-14
Add google-native-ts-k8s-python-postgresql example

adriangb:
This was copied and modified from the `google-native-ts-k8s-ruby-on-rails-postgresql` example.

I think this example is interesting for several reasons.
First off, it uses SQLProxy with Workload Identity IAM, which is the recommended practice as far as I know.
This means that there is no database password stored anywhere and the only user that can connect to the database is the service account.
To my knowledge, there are no public examples of this pattern that document all of the footguns it has (e.g. the `.gserviceaccount.com` suffix removal for the DB user), Pulumi or otherwise.

This also eschews using Gunicorn workers for concurrency in favor of using k8s's HPA, which is also not a pattern that is super well documented.

There's a decent bit of app/Python specific stuff in here, and some of it relates specifically to a web framework I'm working on, but it's just ASGI so users can very easily swap it out for plain Starlette, FastAPI, Blacksheep, etc.
But I'm happy to try to "dumb down" the Python/app part if you'd like


adriangb:
Thanks for running CI! Just FYI it's been a minute since I submitted this, I think I'd like to give it a once over for misc improvements, but also feel free to review it and give me suggestions 😁

guineveresaenger:
@adriangb - I am so sorry for the long wait here. The tests look fine to me - did you want to go ahead and have us merge your example?

guineveresaenger (approved):

adriangb:
Ok just gave this a once over.

guineveresaenger:
Thank you for adding this example! 🎉 Going ahead and merging this in. Have a great week!

adriangb:
Awesome, thank you!

PR #1179, opened 2022-02-23
Fix missing nil return in error condition

guineveresaenger:
Should resolve the following error:https://github.com/pulumi/examples/runs/5285445422?check_suite_focus=true


cnunciato (approved):
Defer to @jkisk, but this looks good to me!

jkisk (approved):
Thank you!

guineveresaenger:
Closing because https://github.com/pulumi/examples/pull/1181 does the same thing

PR #1180, opened 2022-02-23
Add pip install steps to all Python examples

### LINKED ISSUE #1172: Add `pip install` steps to all python examples
Currently, we don't have this step in some of our examples. If you're following the steps exactly (because you're a beginner for example) then you might not realise that you need this.

cnunciato:
This change ensures all Python examples include a `pip install` step, and standardizes on `pip install` for consistency (we seemed to have a mixture of `pip` and `pip3`).

Fixes #1172.

### DIFF
--- a/aws-apigateway-py-routes/README.md
+++ b/aws-apigateway-py-routes/README.md
@@ -26,7 +26,8 @@ For Pulumi examples, we typically start by creating a directory and changing int
 
     ```bash
     python3 -m venv venv
-    venv/bin/pip install -r requirements.txt
+    source venv/bin/activate
+    pip install -r requirements.txt
     ```
 
 2. Create a new Pulumi stack:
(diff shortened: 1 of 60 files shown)


mikhailshilkov:
Are these commands still needed? Could we add `virtualenv: venv` to all Python examples and let the CLI do the venv/pip commands for users? I think this example looks clean: https://github.com/pulumi/examples/tree/master/azure-py-containerapps

lukehoban:
> Could we add virtualenv: venv to all Python examples and let the CLI do the venv/pip commands for users?

Yeah - I actually thought we had already updated everything to use `virtualenv` as part of https://github.com/pulumi/examples/pull/885.

nimbinatus:
Ah, I honestly wasn't aware we had done that, and it's not something a handful of us updating the examples were aware of. I think the issue is the READMEs aren't consistent; and as someone using Python who perhaps hasn't read the docs in depth, it would seem like an odd omission (or if they copy it over from somewhere, such as if they're copying it from the examples repo, the dependencies aren't installed, are they?). Perhaps we should link to https://www.pulumi.com/docs/intro/languages/python/#virtual-environments explicitly in these READMEs to make that clearer? I believe we've been having a few folks saying the system didn't work properly for them, as well, but I don't have the exact details.

lukehoban:
The PR I linked above did try to ensure all readme’s were updated. 

Are we aware of anything concrete that is currently broken?

We could in principle link to those docs, though there shouldn’t be any action the user needs to take related to that - so it would be more of an educational aside. 

> I believe we've been having a few folks saying the system didn't work properly for them, as well, but I don't have the exact details.

Would be good to understand this.

cnunciato:
Thanks for the internal discussion, folks -- I had no idea `virtualenv: venv` was a thing and that it handled all of this automatically. 😄 The docs also make it clear [how to add a Python dependency](https://www.pulumi.com/docs/intro/languages/python/#packages) in this configuration, so I'll close this PR and submit another one for the little tweaks I made alongside the install instructions.

PR #1207, opened 2022-04-21
Mitchgerdisch/canary add bucketversioning

MitchellGerdisch:
Added bucket versioning so that changes to the canary script are propagated to the Canary and underlying lambda function.


phillipedwards (commented):
LGTM

phillipedwards (approved):

PR #1208, opened 2022-04-25
Fix for EKS cluster example

wolfedale:
`*eks.Cluster` has no field or method CertificateAuthority - to make it work we need to change it to `pulumi.StringOutput(eksCluster.CertificateAuthorities)`.


mjeffryes:
Closing as stale since this has not been touched in a few years and no longer applies cleanly.

PR #1210, opened 2022-05-04
Add java examples

t0yv0:
Adds Java examples, tested up to preview, using floating references to Maven Central published artifacts.

The examples should really use Maven not Gradle, I can follow up shortly but perhaps we can check this in?


aws-java-eks-minimal/app/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+repositories {
+    mavenLocal()
+    maven { // The google mirror is less flaky than mavenCentral()
+        url("https://maven-central.storage-download.googleapis.com/maven2/")
+    }
+    mavenCentral()
+}
+
+dependencies {
+    implementation "com.pulumi:pulumi:(,1.0]"
```
pawelprazak, line 14: this might be problematic, we'll se

pawelprazak (approved):

aws-java-eks-minimal/app/build.gradle
t0yv0: We do need a floating reference I think since we want this to use "latest" but perhaps there's a better way to do it?

PR #1216, opened 2022-05-12
readme using stack output substitution

mchaynes:
Adds a stack readme file to stepfunctions example.


EvanBoyle (commented):
@mchaynes can you update the README.md for this example (the one that renders on github) to add some notes about the stack readme and how to access it (go to the console by running `pulumi console` and click on the readme tab).

Can we add a standalone simple example called `ts-readme` with something like:

Pulumi.README.md:
```
# Stack README Title

Full markdown support! Substitute stack outputs dynamically so that links can depend on your infrastructure! Link to dashboards, logs, metrics, and more.

1. Reference a string stack output: ${outputs.strVar}
2. Reference an array stack output: ${outputs.arrVar[1]}
```

```ts
export const strVar = "foo";
export const arrVar = ["fizz", "buzz"];

export const R
```

mchaynes:
@EvanBoyle Updated with more details in the readme

EvanBoyle (approved):
This looks great!!! This will be really helpful for our users to get started. 

Would be great to have examples for each language, but we can save that for another PR.

EvanBoyle (approved):
Looks good to me. Test failures look unrelated. Let's get these shipped!

Thanks for following up with examples for all languages! 🚢

PR #1227, opened 2022-06-05
Fix issue with example role not having access to subresources (such as GSI) etc.

mxkxf:
Fixes [#19](https://github.com/pulumi/pulumi-aws/issues/1987)


stack72 (approved):
Thanks so much @mikefrancis

PR #1228, opened 2022-06-06
Update java examples to maven

t0yv0:


mikhailshilkov (approved):
Not an expert but LGTM directionally

PR #1235, opened 2022-06-16
Add yaml stack readme example

mchaynes:
Adds yaml example for stack readme feature in the pulumi service.

Follow up after `Fn::ReadFile` was added to Pulumi YAML. https://github.com/pulumi/pulumi-yaml/issues/200 (Thanks @AaronFriel!)

Screenshot 2022-06-16 at 14 15 12

Screenshot 2022-06-16 at 14 16 24


EvanBoyle (approved):
🚢

PR #1236, opened 2022-06-17
render readme to string instead of byte array

mchaynes:
instead of returning a byte array, return a string so it renders properly.

```
Outputs:
    arrVar: [
        [0]: "fizz"
        [1]: "buzz"
    ]
    readme: "# Stack README\n\nFull markdown support! Substitute stack outputs dynamically so that links can depend on your infrastructure! Link to dashboards, logs, metrics, and more.\n\n1. Reference a string stack output: ${outputs.strVar}\n2. Reference an array stack output: ${outputs.arrVar[1]}"
    strVar: "foo"

Resources:
    + 1 created

Duration: 1s
```

### DIFF
--- a/stack-readme-ts/README.md
+++ b/stack-readme-ts/README.md
@@ -14,7 +14,7 @@ import { readFileSync } from "fs";
 export const strVar = "foo";
 export const arrVar = ["fizz", "buzz"];
 // add readme to stack outputs. must be named "readme".
-export const readme = readFileSync("./Pulumi.README.md");
+export const readme = readFileSync("./Pulumi.README.md").toString();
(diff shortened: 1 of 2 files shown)


EvanBoyle (commented):
README needs to be updated as well: https://github.com/pulumi/examples/blob/master/stack-readme-ts/README.md?plain=1#L17

mchaynes:
@EvanBoyle Added in https://github.com/pulumi/examples/pull/1236/commits/32eaae5c40f1e4038b2ceed37e2510292b8f0828

EvanBoyle (approved):

PR #1237, opened 2022-06-17
Add an example of using Pulumi with Ansible

joeduffy:
This change adds an example of deploying Wordpress to AWS EC2 using
Pulumi and Ansible together. Pulumi handles provisioning the AWS
infrastructure required to run Wordpress in AWS, including the EC2
instance and an RDS MySQL database, all running within a VPC with
proper public and private subnets, exposed to the Internet using an
Elastic IP address. Ansible handles configuring the EC2 virtual
machine after it's been provisioned with a playbook that knows how
to install and configure Wordpress. The entire deployment is
orchestrated by a single `pulumi up` thanks to our command package.


aws-yaml-ansible-wordpress/Pulumi.yaml
```diff
@@ -0,0 +1,243 @@
...
+  publicKey:
+    Fn::ReadFile: ${publicKeyPath}
+  # Read in the private key for easy use below (and to ensure it's marked a secret!)
+  privateKey:
+    Fn::Secret:
+      Fn::ReadFile: ${privateKeyPath}
+
+resources:
+  # Set up a Virtual Private Cloud to deploy our EC2 instance and RDS datbase into.
+  prod-vpc:
```
jkodroff, line 53: Suggest using AWSX 1.0 for this if it works with YAML (I forget if it does).

jkodroff (approved):
All in all, this looks good.

I'm not the foremost Anisible expert, so I might wait on a second approval for additional comments on how this code might be improved.

aws-yaml-ansible-wordpress/Pulumi.yaml
joeduffy: Thanks Josh. I added a note about the fact that you can choose to use AWSX. Especially for the Ansible audience, I actually thought it would be best to lay out all the raw building blocks so there's a bit less "magic" to the example.

joeduffy:
The Airflow example is failing. This seems to happen from time to time. @stack72 Is this test just flaky? I'll merge anyway but it definitely makes me feel like I'm not doing the right thing 😦

PR #1238, opened 2022-06-17
Fix a lint error

joeduffy:
This seems to have crept in with
https://github.com/pulumi/examples/pull/1233.


cnunciato (approved):
🙀 Sorry I missed this! Thanks for the fix.

PR #1253, opened 2022-07-20
refact: resolve ambigous variable name

gunar:


dixler:
Hey, @gunar could you give me more context for why you want this change? :slightly_smiling_face:

gcp-py-cloudrun-cloudsql/__main__.py
```diff
@@ -19,16 +19,16 @@
     "database", instance=cloud_sql_instance.name, name=config.require("db-name")
 )
 
-users = pulumi_gcp.sql.User(
-    "users",
+user = pulumi_gcp.sql.User(
+    "user",
```
gunar, line 23: It's a single user so no point in being plural.y

gcp-py-cloudrun-cloudsql/__main__.py
```diff
@@ -19,16 +19,16 @@
...
     name=config.require("db-name"),
     instance=cloud_sql_instance.name,
     password=config.require_secret("db-password"),
 )
 
 sql_instance_url = Output.concat(
     "postgres://",
-    config.require("db-name"),
+    user.name,
```
gunar, line 31: Even though `config.require("db-name") === user.name`, it'd be best to reference `user.name` instead so the chain of dependencies is clear to both the reader and to the Pulumi engine.

mjeffryes:

PR #1258, opened 2022-08-10
Update gke java example fixing breaking changes

pawelprazak:
Update GKE Java example to fix imports after breaking changes in Kubernetes provider.

Part of https://github.com/pulumi/pulumi-java/issues/732


t0yv0:
Already on master.

PR #1264, opened 2022-08-16
Unit-testing C# with top-level program + docs

Zaid-Ajaj:
An example that shows how to do unit testing going from a top-level statement that executes the deployment to making assertions in the unit test project about the created resources

Resolves [pulumi#10391](https://github.com/pulumi/pulumi/issues/10391)


testing-unit-cs-top-level-program/README.md
```diff
@@ -0,0 +1,78 @@
...
+In order to unit test this piece of code, we have to do a bit of rewriting such that we don't run the `Deployment.RunAsync` function because it expects the program is being executed from the Pulumi CLI. 
+
+Now, we take the lambda that is provided to `Deployment.RunAsync` and move it in a place where we can reference it later from our unit test project. The following snippet shows this
+
+```csharp
+using Pulumi;
+
+return await Deployment.RunAsync(Deploy.Infra);
+
+class Deploy
```
Frassle, line 23: Does this need to be in a class? Could we reference a top level static method elsewhere?

testing-unit-cs-top-level-program/README.md
```diff
@@ -0,0 +1,78 @@
...
+    public static Dictionary<string, object?> Infra()
+    {
+        // Create resources here
+    }
+}
+```
+Here, the code that creates resources and returns outputs is inside the `Deploy.Infra` function which we can reference from the test project. 
+
+> The return type of `Deploy.Infra` could also have been any of the following 
+> - `void` when we are not returning outputs
```
Frassle, line 34:
```suggestion
> - `void` when we are not returning any outputs
```

testing-unit-cs-top-level-program/README.md
```diff
@@ -0,0 +1,78 @@
...
+> - `Task` when we want to `await` async code
+> - `Task<Dictionary<string, object?>>` when we want to `await` async code _and_ return outputs
+
+There is a catch before we can reference `Deploy.Infra` from the unit tests. Since we don't have an explicit namespace here, we need to add this property to the project file
```
Frassle, line 38: eh this isn't due to namespace, this is because the default visibility for a class is "internal". It might be simpler to make the Deploy class public rather than getting into project changes.

testing-unit-cs-top-level-program/README.md
```diff
@@ -0,0 +1,78 @@
...
+  <InternalsVisibleTo Include="Tests"/>
+</ItemGroup>
+```
+Now we are ready to write unit tests. From a test project called `Tests`, add a reference to the project that contains the Pulumi program:
+```xml
+<ItemGroup>
+  <ProjectReference Include="..\infra\Infra.csproj" />
+</ItemGroup>
+```
+Now, only for the tests we can create a _test stack_ that uses `Deploy.Infra` as follows
```
Frassle, line 50:
```suggestion
Now, just for the tests we can create a _test stack_ that uses `Deploy.Infra` as follows
```

testing-unit-cs-top-level-program/README.md
```diff
@@ -0,0 +1,78 @@
...
+    public Dictionary<string, object?> Outputs { get; set; }
+}
+```
+Then, we can use this test stack to build resources and make assertions about them:
+```csharp
+[Test]
+public async Task StackExportsWebsiteUrl()
+{
+    var options = new TestOptions { IsPreview = false };
+	var resources = await Pulumi.Deployment.TestAsync<TestStack>(new Mocks(), options);
```
Frassle, line 68: I wonder if we should have a TestAsync method that takes the same delegates that RunAsnyc takes? Would save the dummy stack class above.

mikhailshilkov: Yeah, +1 to that. `TestStack` is a bit ugly. Also, will it work if `Infra` is async?

testing-unit-cs-top-level-program/README.md
Zaid-Ajaj: I tried not putting the code in a class 😞 but the `public` modifier is not allowed on top-level functions. I believe because their class is compiler-generated. Then I tried using reflection to get the method info and call it but I couldn't retrieve it. I think it is possible but it could messy so for now, a static method in a class is one way to go

testing-unit-cs-top-level-program/README.md
Zaid-Ajaj: I thought the `public` modifier was not allowed in all of the top-level C# file but it was only for top-level functions so indeed I could mark the class `public` and use from the unit test project without making internals visible. Thanks for the nice catch!

testing-unit-cs-top-level-program/README.md
Zaid-Ajaj: > will it work if Infra is async?

Yes, but the user will have to modify the code a bit:
```cs
Outputs = Deploy.Infra().GetAwaiter().GetResult();
```

> I wonder if we should have a TestAsync method that takes the same delegates that RunAsnyc takes? 

Yes, that's the plan. I also would love it if the following code works
```cs
await Deployment.TestAsync(mocks, options, () => 
{
    return Deployment.RunAsync(() => 
    {
        // Create resources here
    });
});
```
Because then, without changing a user's code when they write
```cs
// Program.cs
return await Deployment.RunAsync(() => 
{
    // Create resources here
});
``` 
We could write a utility function for unit tests that dynamically executes the entry point of the assembly being tested _inside_ a `Deployment.TestAsync` call. No need to rewrite the code to make it unit-testable

Frassle (approved):
Looks ok to me, but we should loop back to the core SDK and see about adding some more overloads to TestAsync

PR #1265, opened 2022-08-18
More robust perf matrix

t0yv0:
Port https://github.com/pulumi/templates/pull/326 enhancements here also. While perf matrix is currently green, it used to fail for a few weeks and this fix should help it be more robust to a failure in only one of the benchmarks.

Sample run: https://github.com/pulumi/examples/actions/runs/2884737144


t0yv0:
https://github.com/pulumi/examples/runs/7902906461?check_suite_focus=true test-drive unfortunately waiting for runners, perhaps Actions: degraded https://www.githubstatus.com/ has something to do with it.

RobbieMcKinstry (approved):
Looks good! How do you test your GHA changes? Do you run them locally? (How?)

PR #1270, opened 2022-08-30
Mitchell gerdisch/ansible examples readmes

MitchellGerdisch:
The ansible-wordpress examples README had a couple of typos in it:
- `pulumi config set publicKeyPath ...` and `pulumi config set privateKeyPath...` used values that did not match the `ssh-keygen` example code earlier in the README.
- The `pulumi config set dbPassword ...` used an example Password that is not valid and so the `pulumi up` fails with this error: `* Error creating DB Instance: InvalidParameterValue: The parameter MasterUserPassword is not a valid password. Only printable ASCII characters besides '/', '@', '"', ' ' may be used.`


pierskarsenbarg (approved):
LGTM

caboog (approved):

PR #1271, opened 2022-09-01
Fix perf data inconsistency by prewarming examples benchmarks

t0yv0:
Noticing that aws-ts-s3-folder shows up as FASTER than aws-js-s3-folder, by a large margin, in our perf data. It can't be true and does not repro locally. This is not great. The hypothesis is that since unlike template benchmarks, this setup here does not call `pulumi new`, the issue is that first Pulumi command ends up downloading plugins needed for the test. This skews the data we want to see. 

The fix attempted here is to run a preview before running a full cycle for measurement.

Test run: https://github.com/pulumi/examples/actions/runs/2974913701

### DIFF
--- a/misc/test/performance_test.go
+++ b/misc/test/performance_test.go
@@ -1,3 +1,4 @@
+//go:build Performance || all
 // +build Performance all
 
 package test
@@ -36,8 +37,8 @@ func TestAccAwsGoS3Folder(t *testing.T) {
 			})
 		},
 	}
-	test := getAWSBase(t).With(opts).With(benchmark.ProgramTestOptions())
-	integration.ProgramTest(t, &test)
+	test := getAWSBase(t).With(opts)
+	programTestAsBenchmark(t, benchmark, test)
 }
 
 func TestAccAwsCsS3Folder(t *testing.T) {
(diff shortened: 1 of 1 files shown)


t0yv0:
This makes the job slower, but I think I'm seeing success! I've downloaded the traces from the artifacts in GitHub Actions, and the trace for empty-update on JS and TS are now identical-ish and around 3.5s on total Pulumi Span 🎉

misc/test/performance_test.go
```diff
@@ -177,6 +178,32 @@ func TestManyResources(t *testing.T) {
...
 
+func programTestAsBenchmark(
+	t *testing.T,
+	bench traces.Benchmark,
+	test integration.ProgramTestOptions) {
+
+	// Run preview only to make sure all needed plugins are
+	// downloaded so that these downloads do not skew
+	// measurements.
+	t.Run("prewarm", func(t *testing.T) {
```
t0yv0, line 189: Go question: does this guarantee prewarm will always run before "benchmark"? It seems to work but I was not sure. I did not want to introduce these blocks but ProgramTest does not like running twice in the same testing.T.

dixler (approved):
LGTM

misc/test/performance_test.go
RobbieMcKinstry: I think only because they are not marked as parallel (the linter in pu/pu might complain about this actually).

RobbieMcKinstry: > ProgramTest does not like running twice in the same testing.T.

That's strange. Yeah, I think semantically it doesn't make sense to "prewarm" as a subtest. We can merge -- I don't think this code is super delicate, so if it works, it works. It's okay if it's less than perfect.

RobbieMcKinstry (approved):

misc/test/performance_test.go
t0yv0: It does seem to work on one example, but we can merge and let it run a few days and get data. This is not customer-facing.

PR #1275, opened 2022-09-07
Fix azure-java-appservice-sql-exmaple post Output invoke rollout

t0yv0:
To make `mvn compile` work again.


t0yv0:
Dup of https://github.com/pulumi/examples/pull/1263, apologies.

PR #1276, opened 2022-09-07
Update utils.py

jsia9:
fix invalid apiVersion "client.authentication.k8s.io/v1alpha1" migrate to v1beta1. which causes the example to fail.


dixler (approved):
LGTM thanks!

I was able to deploy the original example. Apparently there was an issue with the  AWS CLI for some users. I'm not sure if you're directly affected by this.
- https://github.com/aws/aws-cli/issues/6920

PR #1278, opened 2022-09-11
Update README.md

palladius:
you changed public_ip to nginx_public_ip.
So I've updated the READMe accordingly.


dixler (approved):
Thanks for the fix! We really appreciate it.

palladius:
keep up with the great work!

And consider Ruby ;)

PR #1281, opened 2022-09-13
Fix aws-java-eks-minimal example

t0yv0:


aws-java-eks-minimal/src/main/java/com/pulumi/example/eksminimal/App.java
```diff
@@ -18,32 +19,61 @@ public static void main(String[] args) {
...
-
-        var cluster = new Cluster("my-cluster", ClusterArgs.builder()
-                .vpcId(vpcIdOutput)
-                .subnetIds(subnetIdsOutput)
+
+        // Export it for visibility via `pulumi stack output`.
+        ctx.export("defaultVpcId", defaultVpcId);
+
+        // Find all the subnets in the default VPC.
+        final var allSubnetIds = Ec2Functions.getSubnets(GetSubnetsArgs.builder()
```
t0yv0, line 32: Perhaps we should provision the subnets instead of finding them in the default VPC, to make the example more self-contained.

aws-java-eks-minimal/src/main/java/com/pulumi/example/eksminimal/App.java
```diff
@@ -18,32 +19,61 @@ public static void main(String[] args) {
...
+                        .filters(GetSubnetsFilterArgs.builder()
```
t0yv0, line 33: getSubnetIds returned deprecation warnings. https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/subnets is recommended to be used instead. Its not obvious from the types but this is how it's to be used.

t0yv0:
Closing as stale.

PR #1288, opened 2022-09-28
Ensure aws-js-s3-folder does not run tsc

t0yv0:
Since we reuse this example as a benchmark, it can be interesting to compare it against aws-ts-s3-folder with the same functionality to estimate TypeScript overhead. However without this switch, it runs TypeScript compilation anyway.


dixler (approved):
LGTM

PR #1296, opened 2022-10-27
Fix parameter origin domain name error

flo-kn:
The value `websiteEndpoint` seems to produce an error on `$ pulumi up` command 🙋‍♂️❌
```
  aws:cloudfront:Distribution (cdn):
    error: 1 error occurred:
        * error creating CloudFront Distribution: InvalidArgument: The parameter Origin DomainName does not refer to a valid S3 bucket.
        status code: 400, request id: 123456-2345-7890-1234-123456789
```

Use `bucketRegionalDomainName` instead!

Think , a longterm solution could be to exchange the legacy origin access identity (OAI) with origin access control (OAC). See [AWS Docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html#migrate-from-oai-to-oac). Lemme know what you think! I could over support on the migration.

### DIFF
--- a/aws-ts-static-website/index.ts
+++ b/aws-ts-static-website/index.ts
@@ -174,7 +174,7 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
     origins: [
         {
             originId: contentBucket.arn,
-            domainName: contentBucket.websiteEndpoint,
+            domainName: contentBucket.bucketRegionalDomainName,
             s3OriginConfig: {
                 originAccessIdentity: originAccessIdentity.cloudfrontAccessIdentityPath,
             },


cnunciato:
Thanks for the submission @flo-kn! We'll take a look. Did you make any additional customizations to the program to produce this error? If so, would you mind sharing what they were? And out of curiosity, which AWS region were you deploying to?

flo-kn:
Hi @cnunciato,

I modified parts of the code for my own purpose, e.g. I migrated to OAC. That's when I bumped into the error. 

However, before submitting this PR, I started again fresh on a whole new environment (pulumi stack) with code from [pulumi/examples master](https://github.com/pulumi/examples/tree/master/aws-ts-static-website) to make sure that the error did not come from some of my modifications.

Here my diff:
```sh

diff --git a/aws-ts-static-website/index.ts b/aws-ts-static-website/index.ts
--- a/aws-ts-static-website/index.ts
+++ b/aws-ts-static-website/index.ts
@@ -174,7 +174,7 @@ const distributionArgs: aws.cloudfront.DistributionArgs = {
     origins: [
         {
             originId: contentBucket.arn,
-            domainName: contentBucket.websiteEndpoint,
+            domainName: contentBucket.bucketRegionalDomainName,
             s3OriginConfig: {
                 originAccessIdentity: originAccessIdentity.cloudfrontAccessIdentityPath,
             },
```

So no customizations, at least not in the source code.

Region in the profile was: `eu-central-1`:

```sh
$ aws configure get region 
eu-central-1
```

mjeffryes:

PR #1298, opened 2022-10-31
example pulumi go ec2 load balancer static web

holywend:
Add example of a static web on EC2 instances with load balancer. also can set number of needed instance with `srvCount` constant in the main.go file

require additional go dependency to convert number to word, but this can easily modified.


mjeffryes:

Issue #1294, opened 2022-10-12
gcp-ts-functions doesn't work


vbo75:
### What happened?

Hi,
I working on a onepage JS app with REST API backend storage. I want to deploy this app in a full serverless GC config with Pulumi. So I'm testing your sample https://www.pulumi.com/blog/simple-serverless-programming-with-google-cloud-functions-and-pulumi/ and https://github.com/pulumi/examples/blob/master/gcp-ts-functions/index.ts

### Steps to reproduce

Here is the the code that I wrote. Because of intempestive errors with Typescript support (error TS2339: Property 'send' does not exist on type 'Response<any, Record<string, any>>'), I changed to Javascript support.

```
var gcp = require("@pulumi/gcp")

const runtime = "nodejs16";
const endpoint = new gcp.cloudfunctions.HttpCallbackFunction("backend", {
    runtime: runtime,
    callback: (req, res) => {
        console.log('Log say Hello');
        res.send('Hello');
    }
});

const invoker = new gcp.cloudfunctions.FunctionIamMember("invoker", {
    project: endpoint.function.project,
    region: endpoint.function.region,
    cloudFunction: endpoint.function.name,
    role: "roles/cloudfunctions.invoker",
    member: "allUsers",
});

exports.url = endpoint.httpsTriggerUrl;
```

### Expected Behavior

$ curl "$(pulumi stack output url)"
=> Hello

$ pulumi logs --follow
=> Log say Hello

### Actual Behavior

Nothing in command line
Nothing in logs

### Output of `pulumi about`

CLI
Version      3.42.0
Go Version   go1.19.1
Go Compiler  gc

Plugins
NAME    VERSION
gcp     6.40.0
nodejs  unknown

Host
OS       darwin
Version  12.6
Arch     x86_64

This project is written in nodejs: executable='/usr/local/bin/node' version='v16.15.0'

Current Stack: dev

TYPE                                                    URN
pulumi:pulumi:Stack                                     urn:pulumi:dev::archimodel-backend::pulumi:pulumi:Stack::archimodel-backend-dev
gcp:cloudfunctions:CallbackFunction                     urn:pulumi:dev::archimodel-backend::gcp:cloudfunctions:CallbackFunction::backend
pulumi:providers:gcp                                    urn:pulumi:dev::archimodel-backend::pulumi:providers:gcp::default_6_40_0
[...]

PR #1300, opened 2022-11-03
Remove gcp.cloudfunctions.HttpCallbackFunction typings

cnunciato:
To unbreak this example, this PR changes the types of the callback parameters in this example to `any` while we investigate https://github.com/pulumi/pulumi-gcp/issues/939.

Fixes https://github.com/pulumi/examples/issues/1294.


scottslowe (approved):
LGTM (I also tested to verify it works)

PR #1301, opened 2022-11-04
Add a Redshift example

cnunciato:
Part of https://github.com/pulumi/marketing/issues/560.

### DIFF
--- a/aws-ts-redshift/index.ts
+++ b/aws-ts-redshift/index.ts
@@ -1,0 +1,87 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+
+const config = new pulumi.Config();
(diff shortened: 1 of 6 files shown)


aws-ts-redshift/index.ts
```diff
@@ -0,0 +1,87 @@
...
+        {
+            policy: JSON.stringify({
+                Version: "2012-10-17",
+                Statement: [
+                    {
+                        Action: [
+                            "s3:Get*", "s3:List*",
+                        ],
+                        Effect: "Allow",
+                        Resource: "*",
```
cnunciato, line 63: This should probably be restricted to the bucket created on line 37 rather than apply to all of S3.

scottslowe (approved):
Approved (but your note about the S3 bucket is noted, we can address this in a follow-up PR)

PR #1305, opened 2022-11-11
Update examples to use config instead of configuration

### LINKED ISSUE #1304: Update examples to use `config` instead of `configuration`
`pulumi convert `requires a bit more refactoring before `config` can be safely used and `configuration` can be fully deprecated.

aq17:
pulumi-yaml `configuration` is being deprecated in favor of the project-level `config` block 
Fixes #1304

### DIFF
--- a/azure-yaml-container-apps/Pulumi.yaml
+++ b/azure-yaml-container-apps/Pulumi.yaml
@@ -1,7 +1,7 @@
 name: azure-container-apps
 runtime: yaml
 description: Azure Container Apps example
-configuration:
+config:
   sqlAdmin:
     type: String
     default: pulumi
@@ -74,7 +74,7 @@ resources:
     properties:
       resourceGroupName: ${resourceGroup.name}
       kubeEnvironmentId: ${kubeEnv}
-      configuration:
+      config:
         ingress:
             external: true
             targetPort: 80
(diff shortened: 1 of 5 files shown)


kubernetes-yaml/Pulumi.yaml
```diff
@@ -1,6 +1,6 @@
 name: simple-kubernetes
 runtime: yaml
-configuration:
+config:
   hostname:
     default: example.com
     type: String
```
iwahbe, line 6: `String` is not a valid type for `config`. It needs to be `string`.

azure-yaml-container-apps/Pulumi.yaml
```diff
@@ -74,7 +74,7 @@ resources:
     properties:
       resourceGroupName: ${resourceGroup.name}
       kubeEnvironmentId: ${kubeEnv}
-      configuration:
+      config:
```
iwahbe, line 77: Regexp based replace would benefit from `^configuration:` to prevent this kind of thing.

iwahbe (changes requested):

iwahbe:
I don't think that https://github.com/pulumi/examples/pull/1305/commits/c71b9f76176a0bb472f75c9a509cd8e474443dd1 made the changes you wanted it to.

aq17:
> c71b9f7

My bad I've been focusing on OR this morning, still need to fix this

iwahbe (approved):

PR #1306, opened 2022-11-18
Change in stack output variable

jasbir84:


jasbir84 (commented):
Change in the variable name in README.md file

dixler (approved):
Thanks for the catch!

PR #1311, opened 2022-11-24
Changes for kubeconfig to kubeConfig

jasbir84:
Changes for kubeconfig to kubeConfig


lukehoban:
The README at https://github.com/pulumi/examples/blob/78219313610809e6a00f5ead0b0b732b9e2bbfba/gcp-java-gke-hello-world/README.md assumes this output is named `kubeconfig` - so we can't change this without also changing that.  

I'm also not certain we need to diverge on naming here - even in TypeScript, which uses the same camelCase naming convention, we name this `kubeconfig` fairly consistently throughout the `examples` repo.

Are there other changes to the `yamlTemplate` necessary for correctness?  If so, do you want to remove the other changes and we can address just those pieces?

mjeffryes:

PR #1315, opened 2022-12-05
Switch Go azure-native examples to new SDK

thomas11:
fixes pulumi/pulumi-azure-native#2058

Tested via `pulumi stack init dev && pulumi config set azure-native:location westus2 && pulumi preview` in each repo.

Also contains an unrelated fix for azure-go-aks-helm where the referenced apache chart had aged out of the Bitnami repo.


t0yv0 (approved):
Brilliant, thanks. Thanks especially for testing notes.

PR #1320, opened 2022-12-07
Higher entropy go n-res benchmark

t0yv0:
Counteract strings.Repeat being eaten by gzip encoding, add entropy.


caseyyh (approved):
always need more chaos!

PR #1322, opened 2022-12-09
work in progress of fixing up the seo for registry + examples repo

susanev:
- makes unique h1s
- mentions language in h1s
- makes unique intro paragraphs, which is what is used in the meta description


scottslowe (approved):
Just one comment/question regarding the introductory sentence/paragraph on some of the changed examples, but no need to block on that. LGTM otherwise.

PR #1327, opened 2022-12-23
Add Redshift Glue ETL examples in TypeScript and Python

cnunciato:
To accompany https://github.com/pulumi/pulumi-hugo/pull/2345.


scottslowe:
It looks like there are some lint failures in the TypeScript files; I'll hold off on approving until we determine if those lint failures need to be addressed.

cnunciato:
Lint issues addressed @scottslowe!

cnunciato:
I went ahead and ran `black` against the Python files as well just to make sure those are good, too. Assuming there aren't any other issues, it'd be good to merge these because of the links at the end of the blog post that point to them.

scottslowe (approved):
LGTM

PR #1344, opened 2023-02-15
A few more test fixes

cnunciato:
I've done my best to keep the fixes organized with commits, but to summarize the majority of the change here:

* Upgrade failing projects to AWSx 1.0 (with minimal refactorings to make that work)
* Give projects unique names (there was a lot of name reuse, which just invites mishap)
* Updating some filter queries (no-longer-supported syntax, apparently)
* Fixing a few actual bugs
* Adding a few more "expect no changes" exceptions
* Dropping the number of DigitalOcean droplets necessary to convey the point (we frequently run afoul of DO's [documented limits](https://docs.digitalocean.com/products/droplets/details/limits/), so this helps a little bit, but still results in the occasional failure when multiple PR tests run concurrently)  

Fixes https://github.com/pulumi/devrel-team/issues/508.
Part of https://github.com/pulumi/devrel-team/issues/504.

### DIFF
--- a/aws-cs-assume-role/assume-role/Pulumi.yaml
+++ b/aws-cs-assume-role/assume-role/Pulumi.yaml
@@ -1,3 +1,3 @@
-name: assume-role
+name: aws-cs-assume-role
 runtime: dotnet
 description: Demonstrate use of AWS AssumeRole Functionality in C#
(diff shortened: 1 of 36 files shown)


scottslowe (approved):
Lots of changes here, but everything looks reasonable. Approved!

PR #1349, opened 2023-02-17
Fix syntax error in smoke-test step

cnunciato:
Fixes https://github.com/pulumi/devrel-team/issues/512. (PR tests will fail until #1344 merges.)


scottslowe (approved):
LGTM

PR #1351, opened 2023-02-17
Tidy up the cron workflow

cnunciato:
Fixes https://github.com/pulumi/devrel-team/issues/513. 

Specifically:

- Puts workflow names and events at the top of the file -- they're easier to read that way.
- Adds a manual `workflow_dispatch` trigger to the workflow so it can be run on demand.
- Removes the `no-latest-cli` and `no-latest-packages` code paths.

These latter two may have been useful once (near as I can tell, they're intended to let you to toggle between dev and stable versions of the CLI and providers), but since (a) this workflow hasn't hasn't worked for a very long time and no one seems to  care and (b) as of now, our requirements only call for the current version of the CLI and whatever providers are being used by each individual project, we can and should just delete these.

### DIFF
--- a/.github/workflows/add-to-project.yaml
+++ b/.github/workflows/add-to-project.yaml
@@ -4,7 +4,6 @@ on:
     types:
       - opened
       - reopened
-
 jobs:
   add-to-project:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


scottslowe (approved):
LGTM

PR #1352, opened 2023-02-17
More workflow syntax fixes

cnunciato:
As with https://github.com/pulumi/examples/pull/1349, this change just fixes a couple of bad workflow expressions.


scottslowe (approved):
LGTM

PR #1406, opened 2023-03-30
Add AWS hub and spoke example

jkodroff:


jkodroff:
@cnunciato Would be good to merge now if possible so I can ensure the link is correct for the blog post if you're ok with it.

gcp-ts-cloudrun/package.json
```diff
@@ -6,6 +6,7 @@
         "@types/node": "^12.0.0"
     },
     "dependencies": {
+        "@pulumi/awsx": "^0.40.0",
```
scottslowe, line 9: Is there a reason for this seemingly-unrelated change? It also seems like a change we would not want to make.

scottslowe (changes requested):
I haven't run the example yet (will do so shortly), but I am curious about the change to `gcp-ts-cloudrun/package.json`?

scottslowe:
@jkodroff Also, we need to be sure we are providing more complete README's for examples like this. Would you mind making your `README.md` conform to what's outlined in `CONTRIBUTING.md`, please? Sorry, don't mean to be difficult, but at some point we have to start enforcing that proper README's are included for every example. Thanks!

scottslowe:
@jkodroff The updated README looks great, thanks! The only outstanding question for me was the weird change to the `gcp-ts-cloudrun` example, what's up with that?

scottslowe (approved):
Approved, thank you!

PR #1416, opened 2023-04-17
Update hub and spoke example.

jkodroff:


aws-py-hub-and-spoke-network/README.md
```diff
@@ -87,6 +87,14 @@ pulumi destroy
 pulumi stack rm
 ```
 
+## Troubleshooting
+
+You may encounter a condition where the security group fails to delete, this may be due to incomplete deletion of VPC endpoints. To fix this condition, perform the following in the AWS console:
```
scottslowe, line 92:
```suggestion
You may encounter a condition where the security group fails to delete. This may be due to incomplete deletion of VPC endpoints. To fix this condition, perform the following in the AWS console:
```

aws-py-hub-and-spoke-network/README.md
```diff
@@ -87,6 +87,14 @@ pulumi destroy
...
+
+1. Delete all VPC endpoints in the VPC that contains the security group that is failing to delete. Wait for the endpoints finish deletion.
```
scottslowe, line 94:
```suggestion
1. Delete all VPC endpoints in the VPC that contains the security group that is failing to delete. Wait for the endpoints to finish deleting.
```

scottslowe (approved):
A couple small suggestions for you to consider, but otherwise LGTM. (Note that I did not personally test the code.)

PR #1424, opened 2023-05-13
Update aws-ts-lambda-thumbnailer to use latest awsx

cnunciato:
Part of https://github.com/pulumi/pulumi-hugo/issues/2790.


scottslowe (approved):

susanev:
@cnunciato i think this broke a link in the blog, can you fix that up?

https://www.pulumi.com/blog/aws-lambda-container-support/
  -> https://github.com/pulumi/examples/blob/master/aws-ts-lambda-thumbnailer/docker-ffmpeg-thumb/Dockerfile
  -> HTTP_404

Issue #1425, opened 2023-05-18
azure-native load balancer, application gateway examples


phillipedwards:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
Currently, no azure native examples for load balancers and application gateways exist. These resources require you to know azure resource IDs up front which is a bit of a pain and most folks are surprised by this.

PR #1426, opened 2023-05-19
Add Azure-Native Load Balancer (Python) Example

phillipedwards:
partially addresses #1425


cnunciato (approved):
Looks great! Thanks for going the extra step and templatizing. ⭐

Issue #1430, opened 2023-05-27
CLI performance metrics examples broken due to AWS S3 security changes


justinvp:
See https://github.com/pulumi/examples/actions/runs/5090579201/jobs/9149574791#step:15:1

The following tests are failing with similar errors:

- TestAccAwsCsS3Folder
- TestAccAwsFsS3Folder
- TestAccAwsGoS3Folder
- TestAccAwsJsS3Folder
- TestAccAwsPyS3Folder
- TestAccAwsTsS3Folder

```
    aws:s3:BucketPolicy (bucketPolicy):
      error: 1 error occurred:
      	* Error putting S3 policy: AccessDenied: Access Denied
      	status code: 403, request id: MV0GQRM7F6H17R9Q, host id: kKyHGbCnBTbCoY5egn0d2tq6NO2iIxh18n0xGx6glx1oxrPjA0vZ/Ekd8BHFgt3B/vzYk5WrOFQ=
```

It looks like these examples need to be updated along the lines of:

- https://github.com/pulumi/pulumi-yaml/pull/451
- https://github.com/pulumi/pulumi-hugo/pull/2719

cnunciato:
Fixed via #1470.

PR #1433, opened 2023-05-27
Bump Go, Node, and Python versions for perf benchmarks

### LINKED ISSUE #1431: CLI performance metrics TestManyResources examples broken
See https://github.com/pulumi/examples/actions/runs/5090579201/jobs/9149574791#step:15:1899

```
go mod tidy: go.mod file indicates go 1.18, but maximum supported version is 1.16
```

justinvp:
Bump Go, Node, and Python versions for perf benchmarks to use the same versions used in other workflows in this repo, e.g.:

https://github.com/pulumi/examples/blob/30527dbd4ef045781873be08ef955b4e66600eac/.github/workflows/cron.yml#L474-L484

Fixes #1431

### DIFF
--- a/.github/workflows/performance_metrics_cron.yml
+++ b/.github/workflows/performance_metrics_cron.yml
@@ -106,10 +106,10 @@ jobs:
         dotnet-version:
           - 3.1.301
         go-version:
-          - 1.16.x
+          - 1.19.x
         node-version:
-          - 14.x
+          - 16.x
         platform:
           - ubuntu-latest
         python-version:
-          - "3.7"
+          - "3.9"


.github/workflows/performance_metrics_cron.yml
```diff
@@ -106,10 +106,10 @@ jobs:
         dotnet-version:
           - 3.1.301
```
justinvp, line 107: Note: It looks like this will need to be bumped to .NET 6 as part of fixing `TestAccAwsCsS3Folder` and `TestAccAwsFsS3Folder` for #1430 due to:

https://github.com/pulumi/examples/actions/runs/5090579201/jobs/9149574791#step:15:77

```
error NU1202: Package Pulumi.Aws 5.41.0 is not compatible with netcoreapp3.1 (.NETCoreApp,Version=v3.1). Package Pulumi.Aws 5.41.0 supports: net6.0 (.NETCoreApp,Version=v6.0)
```

I'll do that as part of #1430

cnunciato (approved):

Issue #1438, opened 2023-06-13
Update examples to use Docker v4


AaronFriel:
## Issue details

As with pulumi/templates#567, we should update examples to use the latest Docker provider.
```[tasklist]
### Tasks
- [ ] https://github.com/pulumi/examples/issues/1591
- [ ] https://github.com/pulumi/examples/issues/1592
- [ ] https://github.com/pulumi/examples/issues/1593
- [ ] https://github.com/pulumi/examples/issues/1594
```

scottslowe:
The following TypeScript examples appear to be affected:

`azure-ts-containerapps`
`classic-azure-ts-cosmosapp-component`
`gcp-ts-docker-grc-cloudrun`
`docker-ts-multi-container-app`
`gcp-ts-cloudrun`
`classic-azure-ts-aks-keda`
`google-native-ts-k8s-python-postgresql`
`gcp-ts-k8s-ruby-on-rails-postgresql`
`google-native-ts-k8s-ruby-on-rails-postgresql`
`classic-azure-ts-appservice-springboot`
`azure-ts-appservice-docker`
`aws-ts-hello-fargate`

The following Go examples seem to be affected:

`azure-go-appservice-docker`
`azure-go-containerapps`

The following Python examples seem to be affected:

`docker-py-multi-container-app`
`azure-py-containerapps`
`azure-py-appservice`
`aws-py-voting-app`
`aws-py-django-voting-app`

This list is not necessarily complete, but it is a reasonable starting point.

scottslowe:
It's also unclear to me how https://github.com/pulumi/pulumi-awsx/issues/1010 may affect this

rshade:
This has been completed.

Issue #971, opened 2021-04-16
Audit examples for default AWS VPC assumptions


joeduffy:
Some of our AWS examples assume you have a default VPC, and break if you do not. It is often common security practice to delete an account's default VPC upon creation to avoid accidental ingress/egress exposure of private endpoints. And us-east-1 accounts created before a certain date don't have one. I don't have the exact failure information but speaking as someone who had such an account, the failure mode was not pleasant. We should audit our examples for those that depend on default VPCs and call this out in the instructions and/or code as a comment. I wish the error here was much better too.

mikemaccana:
Just ran into this bug now - how would you advise a potential customer to work around this and get the examples running?

```
~/Code/mycompany/pulumi-examples/aws-ts-hello-fargate$ pulumi stack init dev
Created stack 'dev'
~/Code/mycompany/pulumi-examples/aws-ts-hello-fargate$ pulumi config set aws:region us-east-2
~/Code/mycompany/pulumi-examples/aws-ts-hello-fargate$ pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/mikemaccana/aws-ts-hello-fargate/dev/previews/65dad80e-a460-47a9-8121-d58c9345f580

     Type                 Name                      Plan     Info
     pulumi:pulumi:Stack  aws-ts-hello-fargate-dev           1 message

Diagnostics:
  pulumi:pulumi:Stack (aws-ts-hello-fargate-dev):
    error: It looks like the Pulumi SDK has not been installed. Have you run npm install or yarn install?

error: failed to load language plugin nodejs: could not read plugin [/home/mike/.pulumi/bin/pulumi-language-nodejs] stdout: EOF
~/Code/mycompany/pulumi-examples/aws-ts-hello-fargate$ npm i
npm WARN deprecated uuid@3.3.2: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.

added 119 packages, and audited 120 packages in 19s

26 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
~/Code/mycompany/pulumi-examples/aws-ts-hello-fargate$ pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/mikemaccana/aws-ts-hello-fargate/dev/previews/6994f5e3-0e40-41d6-bfca-9567881c64a3

     Type                                          Name                        Plan       Info
[...]

PR #1440, opened 2023-06-20
Fix Go example for AWS Fargate

scottslowe:
This PR improves the Go example for AWS Fargate with a number of changes:

* Remove dependency on default VPC (part of #971)
* Switch to Docker 4.x provider (part of #1438)
* Fix issue when building from ARM64-based systems
* Improve comments


cnunciato (approved):
Looks good!

PR #1442, opened 2023-06-23
Mitchgerdisch/fix integration test program

MitchellGerdisch:
The Pulumi program being tested was using an old version of the aws provider and needed another resource to allow the policy to be added to the bucket. So updated that.
Also there were some deprecation warnings related to some of the go libraries. So updated that.


scottslowe (approved):
These changes seem reasonable, approved. Double-check the failures in the checks to be sure they aren't caused by your changes.

MitchellGerdisch:
Thanks for the review.
None of the failed checks are related.

PR #1470, opened 2023-07-28
Fixes to unblock cron tests

lukehoban:
Fixes https://github.com/pulumi/examples/issues/1463.


lukehoban:
First green test run in pulumi/examples since March 2022 🎉 

https://github.com/pulumi/examples/actions/runs/5688931861/job/15420275443?pr=1470

cnunciato (approved):

Issue #1466, opened 2023-07-21
Update kubernetes examples to use v4


mnlumi:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Update kubernetes examples to use v4 to reflect the latest version of the provider.

mnlumi:
As a result of https://github.com/pulumi/pulumi-kubernetes/issues/2362 being shipped.

PR #1472, opened 2023-08-03
Update programs to kubernetes 4.0.3

mnlumi:
Resolves https://github.com/pulumi/examples/issues/1466


mnlumi:
Opened https://github.com/pulumi/pulumi-kubernetes/issues/2523

mnlumi:
@cnunciato Could you take a look at `kubernetes-go-helm-wordpress`? helmv2 has been removed from kubernetes v4, but updating to helmv3 and the chart version of wordpress is failing with an error.

```
 pulumi:pulumi:Stack (kubernetes-go-helm-wordpress-dev):
    panic: interface conversion: interface {} is nil, not *v1.Service
    goroutine 89 [running]:
    main.main.func1.1({0x0?, 0x0?})
```

kubernetes-go-helm-release-wordpress/go.mod
```diff
@@ -1,57 +1,72 @@
 module github.com/pulumi/examples/kubernetes-go-helm-release-wordpress
 
-go 1.17
+go 1.18
```
desteves, line 3: Do we want to bump this even more to the latest stable `1.21` ?

desteves (commented):
The `go` version is inconsistent and quite old `1.13` at times. Looks like it was changed in some places. I'd suggest to bump them all to the latest stable version

scottslowe:
Bumping this to M96 due to Helm v3 issue described above

scottslowe:
I think I have working code for this. It works as expected on EKS (after I fixed a problem with the EKS template); now testing on GKE.

mnlumi:
@scottslowe At this point, the PR is now several releases behind for Kubernetes. My suggestion is to exclude the broken example while opening an issue to handle it separately.

scottslowe:
@mnlumi I'm good with that approach, assuming no one else has objections.

scottslowe:
@mnlumi Per Eric's e-mail, we'll leave this issue open to track resolution of the broken example. Just exclude the broken example from the PR to update the provider for all the rest of the examples. Once we have all tests passing, then we can merge that PR to get everything else updated.

mnlumi:
Closing as this as it is significantly out of date and a new pull request would be more efficient. @EronWright @mjeffryes  - You may want to work with your team or @cnunciato to get these kubernetes examples updated to v4 instead of v3 and the latest pulumi versions as well.

Issue #779, opened 2020-08-28
Provide example of using Pulumi to deploy NextJS app on AWS


thesobercoder:
Please provide an example of how to use Pulumi to deploy NextJS to AWS replicating how deployments work on Vercel infrastructure with edge network.

I'm looking for an example which replicates the same behavior as this [component](https://github.com/serverless-nextjs/serverless-next.js) does for `serverless`.

vha14:
I'm giving this a crack, starting with https://github.com/dealmore/terraform-aws-next-js

tlinhart:
@thesobercoder According to [this issue](https://github.com/serverless-nextjs/serverless-next.js/issues/777), there's a chance that v2 will come closer to what you (and me as well) desire.

gunar:
A basic example—onto which we'd build later on—might be:

- Serving static content from CloudFront
- Deploying all api routes as a single Lambda

In theory, `next build` provides all things necessary for this to work.

blackgirlbytes:
for anyone struggling to find the answer to this like I was..I attempted to upload this to an S3 bucket, but struggled due to lack of information. 

What worked for me: 
- I ran `npm build && npm export` which generated the static HTML files for me.
- Then I ran aws s3 sync ../{folder path}/ s3://$(pulumi stack output bucketName) --acl public-read

expelledboy:
I can not find (and trust I have looked) any working example with pulumi. However, I have [found a template](https://github.com/serverless-nextjs/serverless-next.js/blob/master/packages/serverless-components/nextjs-cdk-construct/src/index.ts) to start from to begin the process. Scary stuff...

T04435:
Any updates on this, using Pulumi not any other infra as code provider?

Thanks.

AutomationD:
How critical is the advantage of Edge-line deployment vs EC2/ECS + Cloudfront / Cloudflare?

lukehoban:
The PR at https://github.com/pulumi/examples/pull/1475 includes an example of a component that deploys the infrastructure needed to deploy a high-performance Next.js site on AWS using Pulumi.  This leverages [OpenNext](https://open-next.js.org/) which providers a more robust way to package Next.js builds for AWS (and AWS serverless specifically).

AutomationD:
Thanks @lukehoban , looks good!

I'm curious, what are the main benefits of OpenNext deployment vs a container behind a CDN and how important are they for the average user?

PR #1475, opened 2023-08-07
Add a Next.js Site example

lukehoban:
Adds an example of deploying a Next.js site with Pulumi, building on top of [OpenNext](https://open-next.js.org/).

Fixes https://github.com/pulumi/examples/issues/779.


filipef101:
Amazing, just sharing my experimentations with this:
Notes:
 - Not sure if needed the npm install & open next build step, IMO that should be handled by the user or CI before triggering the deployment, or it should be generic, ie "build command".
 - Doesn't use/deploy the warmer lambda
Did play modifying this to deploy to lambda@edge
Notes:
 - Doesn't use/deploy the revalidation lambda function
 - Doesn't setup cloudwatch (not sure if edge specific or not), terraform examples I've seen do mention something about cloudwatch.
 
What I've learned about making @edge lambdas, since I couldn't find anything on Pulumi docs, to deploy a lambda to edge the steps are: 

- lambda needs to be on us-east-1 region
```
ie: just deploy everything to us-east-1, or pass the provider only to the edge lambda (didn't try)
const useast1 = new aws.Provider('useast1', { region: 'us-east-1' });
const site = new NextJsSite(_prefixWithAppName('mysite'),
  { path: '../' },
  { provider: useast1 },
);
```
- adding the lambdaFunctionAssociations on the/a cloudfront behaviour 
   - where the arn is <function>.qualifiedArn
- on the lambda enable "publish" flag
- serverFunction role needs to include edgelambda role:
` ... Service: ['edgelambda.amazonaws.com', 'lambda.amazonaws.com'] ...`
- lambda arch needs to be `x86_64`
- lambda cant take any env variables. (not sure if there's a use case for this) I only changed the server lambda to run on edge so not sure if image lambda can be deployed to edge as you are adding BUCKET_NAME & BUCKET_KEY_PREFIX env vars.

Looks kind of simple to support edge deployments, the terraform example I've seen does it based on a flag using different settings based on it.

I've looked at https://github.com/RJPearson94/terraform-aws-open-next for context
[lambda](https://github.com/RJPearson94/terraform-aws-open-next/blob/main/modules/tf-aws-lambda/main.tf)
[main](https://github.com/RJPearson94/terraform-aws-open-next/blob/main/main.tf)

cnunciato (approved):
Nice! Followed the README and used the Deploy button, and it worked like a charm (after changing the project name). 🚀

cnunciato:
Test failures are unrelated, so I'll go ahead and merge this. Thanks @lukehoban!

@filipef101 Excellent feedback, thanks for taking the time! I'll file a follow-up issue so these suggestions don't get lost.

PR #1481, opened 2023-08-22
Snowflake example

jkodroff:


scottslowe (approved):

snowflake-py-pulumi-search-export/Pulumi.yaml
```diff
@@ -0,0 +1,6 @@
+name: snowflake-pulumi-search-export-py
+runtime:
+  name: python
+  options:
+    virtualenv: venv
+description: A minimal Python Pulumi program
```
cnunciato, line 6: Should this be something like "An example of exporting Pulumi Cloud data to Snowflake"?

snowflake-py-pulumi-search-export/__main__.py
```diff
@@ -0,0 +1,352 @@
...
+    )
+)
+
+function = aws.lambda_.Function(
+    lambda_name,
+    aws.lambda_.FunctionArgs(
+        role=lambda_role.arn,
+        runtime="python3.8",
+        handler="handler.handle",
+        code=pulumi.FileArchive('./.build'),
```
cnunciato, line 103: This seems to fail on `pulumi up`: 

```
Exception: failed to compute archive hash for "code": couldn't read archive path './.build': stat ./.build: no such file or directory
```

Should it be this?

```suggestion
        code=pulumi.FileArchive('./search_export_to_s3'),
```

snowflake-py-pulumi-search-export/__main__.py
```diff
@@ -0,0 +1,352 @@
...
+import json
+import os
+
+bucket = aws.s3.Bucket("pulumi-search-export")
+
+config = pulumi.Config()
+# NOTE: In a production scenario, for security reasons this should should call
+# `config.require_secret()`, use a Pulumi access token specifically designated
+# for this export process, and not fall back to an environment variable:
+pulumi_access_token = config.get_secret(
```
cnunciato, line 18: Preview fails without a `PULUMI_ACCESS_TOKEN`. Should we document this in the README as a pre-req?

cnunciato (changes requested):
Looks like a couple lil' things need fixing up, but otherwise looks great!

snowflake-py-pulumi-search-export/__main__.py
jkodroff: The diagnosis is correct, but the issue is that `.build/` is not properly created in all environments. The Lambda code has to be staged in `.build/` because of how Python Lambdas are packaged. Will fix.

snowflake-py-pulumi-search-export/__main__.py
jkodroff: Yes. Will fix.

jkodroff:
We're probably changing the approach for the Snowflake blog post b/c Pulumi Cloud data export is an Enterprise+ and is not easy to seed with data. I'll come back and finish this PR once we have the code that will be used in the blog post finished.

scottslowe:
@jkodroff I've bumped this from 0.93 to 0.94. If you don't think you'll wrap it up in this sprint, then feel free to adjust the milestone accordingly. Thanks!

PR #1489, opened 2023-08-28
Fix aws-ts-containers compilation

t0yv0:
It appears that "name" property is required. This currently causes CI failure to compile typescript for this example.

https://github.com/pulumi/pulumi-awsx/blob/master/awsx/schema.json#L1208 is the reference.

https://github.com/pulumi/pulumi-awsx/pull/1069 is the change that made this required. 

CC @rquitales @mjeffryes

### DIFF
--- a/aws-ts-containers/index.ts
+++ b/aws-ts-containers/index.ts
@@ -27,6 +27,7 @@ const service = new awsx.ecs.FargateService("service", {
     assignPublicIp: true,
     taskDefinitionArgs: {
         container: {
+            name: "service-container",
             image: image.imageUri,
             cpu: 128,
             memory: 512,


mjeffryes:
We've broken examples a few times recently when updating providers; We should probably be building examples when making providers releases? Alternatively, should the examples versions be pinned somehow and updated after the release goes out?

mjeffryes (approved):
Approving b/c this looks correct & I don't think the check failures are related to this change, but I can't claim to fully understand the failures

t0yv0:
> We should probably be building examples when making providers releases

Well it's a possibility; it's going to be a draw on compute time but can help maintain this repo for sure.

The check failures are present on master independently yes.

PR #1502, opened 2023-09-22
Update aws, awsx, eks, and friends

thomas11:
Update all examples but five to latest aws, awsx, eks, and aws-apigateway. **This update is to the latest alpha versions directly - only for testing, not for merging**.

Also updated a few other dependencies along the way when I came across them.

I haven't fully figured out how to handle the breaking changes in these upgrades:
  - aws-ts-apigateway-auth0
  - aws-ts-lambda-efs
  - aws-ts-netlify-cms-and-oauth/cms-oauth/infrastructure
  - aws-ts-pulumi-webhooks
  - aws-ts-slackbot


mnlumi:
@thomas11 The PR looks good, but I can't see why the one check is failing on destroy. Is this unrelated?

aws-py-hub-and-spoke-network/requirements.txt
```diff
@@ -1,3 +1,3 @@
 pulumi>=3.0.0,<4.0.0
-pulumi_aws>=5.0.0,<6.0.0
-pulumi_awsx>=1.0.0,<2.0.0
+pulumi_aws>=6.0.0,<7.0.0
+pulumi_awsx==2.0.0b3
```
scottslowe, line 3: Is there a reason we're specifying what appears to be a beta release here? And pinning the version?

scottslowe (changes requested):

aws-py-hub-and-spoke-network/requirements.txt
thomas11: An oversight, thanks!

scottslowe (approved):
Approved, thanks!

PR #1507, opened 2023-10-10
Create Pulumi to AWS OIDC Configuration Example

toriancrane:
This example will create an automated way to deploy the OIDC configuration between Pulumi and AWS. 

Will resolve [#3481](https://github.com/pulumi/pulumi-hugo/issues/3481)


aws-py-oidc-provider/__main__.py
```diff
@@ -0,0 +1,70 @@
+import pulumi
+from pulumi_aws import iam
+import requests
+import subprocess
+import OpenSSL
```
jkodroff, line 5: You can use the TLS provider to generate certs: https://www.pulumi.com/registry/packages/tls/

aws-py-oidc-provider/__main__.py
```diff
@@ -0,0 +1,70 @@
...
+import json
+
+audience = "" # Provide the name of your Pulumi Organization
```
jkodroff, line 8: Use `config.require()`: https://www.pulumi.com/docs/reference/pkg/python/pulumi/#pulumi.Config.require

aws-py-oidc-provider/__main__.py
```diff
@@ -0,0 +1,70 @@
...
+oidc_idp_url = 'https://api.pulumi.com/oidc'
+base_url = 'api.pulumi.com/oidc'
+
+# Obtain the OIDC IdP URL and form the configuration document URL
+print("Forming configuration document URL...")
```
jkodroff, line 13: You probably don't want `print` statements because they will print even if you run the program a second time and nothing changes.

aws-py-oidc-provider/__main__.py
```diff
@@ -0,0 +1,70 @@
...
+configuration_url = f'{oidc_idp_url}/.well-known/openid-configuration'
+
+# Locate "jwks_uri" and extract the domain name
+print("Extracting domain name from jwks_uri...")
```
jkodroff, line 17: This should probably be modeled as a resource using the Command provider: https://www.pulumi.com/registry/packages/command/api-docs/local/command/

jkodroff:
Suggest using "Pulumi Cloud" in the docs and `pulumi-cloud` in the code/for the example name.

aws-py-oidc-provider/__main__.py
toriancrane: The steps that use OpenSSL are obtaining Pulumi's OIDC IdP certificate chain and using that to produce its thumbprint (it's not creating a new one). It wasn't clear to me how pulumi_tls would recreate this process, so happy to pair with you on this!

aws-py-oidc-provider/__main__.py
toriancrane: Is there a built-in way to retrieve the org associated with the stack? Or do users have to explicitly configure that?

aws-py-oidc-provider/__main__.py
toriancrane: Is there a Pulumi-specific way of doing this to where it doesn't print if there's been no changes? I tried with `pulumi.log.info()` and it still does it. I don't want to remove the statements because I think the indicators are helpful for users.

aws-py-oidc-provider/__main__.py
toriancrane: I don't really understand the ask here. Can we sync on this and the TLS comment?

aws-py-oidc-provider/__main__.py
scottslowe: There's a built-in way to retrieve the organization, project, and stack name. I've done this in Go many times, but in Python I think you need to use `pulumi.get_organization`. I'll try to find a link.

PR #1509, opened 2023-10-11
Create Pulumi to Azure OIDC Configuration Example

### LINKED ISSUE #1510: Content Enhancement - Add Pulumi-Azure OIDC Example for Pulumi ESC
## Description

To make use of the provider integrations in Pulumi ESC, users need to create an OIDC provider in their chosen cloud. This process is not exactly the most straightforward. This content addition will create a Pulumi program that will automate the creation of an OIDC provider in Azure and provide it as an example in the Examples repo.

toriancrane:
Create Pulumi to Azure OIDC Configuration Example

Will resolve #1510


scottslowe:
@toriancrane Do you know if we're able to test this with our Pulumi Azure subscription? Sounds like as long as I have a unique Entra application/service principal then I should be able to test this.

toriancrane:
> @toriancrane Do you know if we're able to test this with our Pulumi Azure subscription? Sounds like as long as I have a unique Entra application/service principal then I should be able to test this.

Yes, unlike AWS you can make it multiple times in Azure. The issuer is specified in the Federated credentials which seems to be unique per app.

azure-py-oidc-provider-pulumi-cloud/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+    ```bash
+    pulumi config set environmentName <your-environment-name> # replace with your environment name
+    pulumi config set azure-native:location WestUS2 # any valid Azure region will work
+    ```
+
+1. Install requirements.
+
+    ```bash
+    python -m venv venv
```
desteves, line 40: I had to do python3

scottslowe (approved):
Good job; it worked without any issues.

PR #1516, opened 2023-10-31
add performance test including policy packs

tgummerer:
Add a performance test that includes a policy pack, so we can monitor for any regressions over time.

Iiuc we can then add an alert for this on metabase?  Or am I mistaken on that?

### DIFF
--- a/misc/test/performance_test.go
+++ b/misc/test/performance_test.go
@@ -1,12 +1,10 @@
-//go:build Performance || all
-// +build Performance all
-
 package test
 
 import (
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
 	"path"
 	"testing"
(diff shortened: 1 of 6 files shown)


misc/test/performance_test.go
```diff
@@ -1,12 +1,10 @@
-//go:build Performance || all
-// +build Performance all
-
```
justinvp, line 3: I'm not familiar with how we build for the perf tests in this repo, but I assume we want to keep these.

justinvp (approved):
LGTM otherwise

misc/test/performance_test.go
tgummerer: Yes indeed.  Thanks for catching this!

PR #1519, opened 2023-11-07
Quality of life updates for Azure OIDC Python example

toriancrane:
This PR updates the Azure Python OIDC example with the following:

- Incorporates `environmentVariables` needed to run `pulumi up` with dynamic credentials
- Removes resource group definition
- Adds a role assignment as part of the Python code rather than as an example in text
- Updated readme to remove the role assignment content

Inspired by [this gist](https://gist.github.com/MitchellGerdisch/09a6fa382f9996adb25cab13f4c5a166).


scottslowe (approved):
LGTM

PR #1522, opened 2023-11-09
Updates example output in validate step

toriancrane:
This PR updates the example output in the validate step to reflect the latest version.


scottslowe (approved):
Approved, but I have a stupid question: the example YAML output by the Pulumi Python program doesn't include the `environmentVariables` section, so why are we adding it to the README? I'm sure I'm just missing something.

PR #1528, opened 2023-11-17
Adds note about Sub ID bug

toriancrane:
This PR adds a note about the [known subject identifier bug](https://github.com/pulumi/pulumi/issues/14509).

### DIFF
--- a/aws-py-oidc-provider-pulumi-cloud/README.md
+++ b/aws-py-oidc-provider-pulumi-cloud/README.md
@@ -93,6 +93,19 @@ You can configure more granular access control by adding the `sub` claim to the 
     ]
 }
 ```
+
+{{< notes type="warning" >}}
+
+If you are integrating Pulumi ESC with Pulumi IaC, using the specific name of the ESC environment in the subject identifier will not work at this time. There is a [known issue](https://github.com/pulumi/pulumi/issues/14509) with the value of the subject identifier that is sent to the provider from Pulumi.
+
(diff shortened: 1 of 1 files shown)


aws-py-oidc-provider-pulumi-cloud/README.md
```diff
@@ -93,6 +93,19 @@ You can configure more granular access control by adding the `sub` claim to the
     ]
 }
 ```
+
+{{< notes type="warning" >}}
```
cnunciato, line 97: This a Hugo shortcode expression, so it won't render properly outside of our website (e.g., in this GitHub repo). We should adjust it to be regular GitHub-flavored Markdown.

cnunciato (changes requested):

aws-py-oidc-provider-pulumi-cloud/README.md
toriancrane: Hey @cnunciato,

Doesn't this get funneled into the registry? I think this is why I included it if I remember correctly. What's the best way to highlight this in both places?

cnunciato: Sorry for the delayed response! I completely missed this.

It does, but we want it to render properly on GitHub also. Right now, it looks like this:

image

Best thing would probably be to use a regular heading and paragraph to help call attention to it. Folks sometimes use blockquotes for callouts as well (`>`), so that's another option. Either would be fine, IMO.

mjeffryes:

PR #1535, opened 2023-12-27
use pulumi  CLI dev version for running CI

tgummerer:
We recently started publishing dev versions of the pulumi CLI. Use these for the CI builds in this repository, to give it some early exposure, and hopefully catch some bugs early.


.github/workflows/cron.yml
```diff
@@ -43,7 +43,9 @@ jobs:
       with:
         dotnet-version: ${{matrix.dotnetversion}}
     - name: Install Pulumi CLI
-      uses: pulumi/actions@v4
+      uses: pulumi/actions@5ffa6249eba838c2a624d5315a28fb8c7b3dabb7
```
Frassle, line 46: Do we really want this pinned? Isn't latest v4 always going to be fine for us?

tgummerer: It's pinned just to get the latest changes from `pulumi/actions`.  We should probably do a release there so the latest v4 includes the changes to allow us to install the dev version.

Frassle (approved):

Frassle:
We should also update https://github.com/pulumi/home/wiki/Producing-the-Pulumi-CLI-and-SDKs to say check https://github.com/pulumi/examples/actions/workflows/cron.yml is all green. (And work out why it's not green now)

Frassle (changes requested):
We'll need to bump actions to v5.

Frassle (approved):

cnunciato (changes requested):
I'd prefer that we not have this repo take a dependency on dev CLIs -- at least not without more discussion. If we want to use the examples in this repo for testing dev CLIs, we totally can, but we should do so with a separate workflow that's probably managed outside of this repo (e.g., cloning this repo for a pulumi/pulumi job would probably give you what you need).

tgummerer:
@cnunciato happy to discuss this more!  The idea here is to enroll repos in the pulumi org to use the dev version of pulumi/pulumi for running CI, so we can catch any potential user facing issues before they are getting released.  It is especially helpful to this in this repository since it's one of the biggest collection of pulumi code.  

I would really like to keep these changes at least in the `cron` action.  Having a separate job for the dev version there would just mean extra CI time used for running the examples with the same released pulumi version more often.  Especially in light of our already increased CI spend (https://github.com/pulumi/home/issues/3244), I would like to avoid adding additional jobs when we can avoid that.

Note that I wouldn't expect "dev" versions to be much less stable than release versions currently are.  They all go through the usual automated testing and review process before being merged and becoming a dev version.  It's also what we cut a release from currently, and having an early indication of when anything is broken with our current build would be super helpful.  This would also mean that the IaC Core team would start paying closer attention to these CI jobs, which would further help with the stability of them as well.

All this said, I'd be happy to discuss this on slack or zoom as well if you prefer.

cnunciato (approved):
Thanks again for the discussion @tgummerer!

PR #1536, opened 2024-01-02
upgrade pulumi/pulumi dependencies

tgummerer:
This repo is currently using old pulumi/pkg and pulumi/sdk dependencies.  These depend on the no longer existing appdash-data package, so the tests are currently broken in CI.

Update the dependencies to unbreak CI.


tgummerer:
pulumi-trace-tool also need its dependencies updated, so I'm converting this to a draft for now.

tgummerer:
Turns out we still need to wait for the pulumi 3.100 release, as 3.99 still has a dependency on pulumi-java that has an older dependency on pulumi/pulumi.

cnunciato:
@tgummerer I went ahead and addressed this with #1538 to unblock the tests, so I think we can close this for now. Thanks for submitting though!

tgummerer:
Great, thanks!

PR #1539, opened 2024-01-04
Install gotestfmt for the cron test

cnunciato:
This job depends on `gotestfmt`, but it [looks like we aren't installing it](https://github.com/pulumi/examples/actions/runs/7416039140/job/20180271977). This should fix that. 🤞


cnunciato:
Cool, looks like this indeed gets the failing AWS tests to pass. ✅

scottslowe (approved):

PR #1569, opened 2024-01-30
Update deploy with pulumi button to be more visible in dark mode

zchase:
In dark mode the "Deploy with Pulumi" button can be hard to see:

deploy-with-pulumi-darkmode

This PR update the button to include a dark mode version:

deploywithpulumilight

I doubled checked the tutorial generation for the website and these images are not used so we should not have to deal with two buttons rendering on those pages.

Fixes: https://github.com/pulumi/marketing/issues/932
Part of: https://github.com/pulumi/marketing/issues/951


scottslowe (approved):
I reviewed the first 124 files in this PR and didn't spot any errors or issues, so I'm approving the PR.

PR #1576, opened 2024-02-06
Update providers during perf measure

t0yv0:
For Python and TypeScript, provider references are floating, but for Go they are always pinned. As a result TestAccAwsGoS3Folder was always testing the performance of aws v6.0.2 not the later versions that shipped some performance improvements. Compensate for this in the benchmark by updating to latest.

### DIFF
--- a/misc/test/performance_test.go
+++ b/misc/test/performance_test.go
@@ -1,5 +1,3 @@
-//go:build Performance || all
-// +build Performance all
 
 package test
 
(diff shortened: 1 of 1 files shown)


cnunciato (approved):

t0yv0:
Test failure appears unrelated, I think I can merge. I'll revert if something breaks in the metrics pipeline per se.

PR #1580, opened 2024-02-09
fix titles for CI jobs; use dev version for perf tests

tgummerer:
Correct the title for the install actions, as we're now using the dev version of pulumi in the cron jobs.

In doing this I also noticed we are not using the dev version for performance testing, though we should to get ahead of performance regressions before releasing the CLI.

/cc @justinvp who noticed this the other day.


PollRobots (approved):
LGTM

PR #1584, opened 2024-02-12
Do not run benchmark prewarm in parallel with the benchmark

t0yv0:
Looks like ProgramTest defaults to enabling t.Parallel(), and this defeats the purpose of prewarm phase in provider benchmarks. The intent of prewarm was to do a no-op to make sure dependencies are downloaded and avoid measuring that overhead as part of the benchmark. This was not the case though because prewarm was running in parallel with the actual benchmark.

After the change, each individual benchmark first does unmeasured prewarm, and then does the measurement. In the future we may want to measure prewarm as it is interesting for user experience as well.

### DIFF
--- a/misc/test/performance_test.go
+++ b/misc/test/performance_test.go
@@ -230,6 +230,7 @@ func programTestAsBenchmark(
 			SkipUpdate:               true,
 			AllowEmptyPreviewChanges: true,
 			AllowEmptyUpdateChanges:  true,
+			NoParallel:               true,
 		})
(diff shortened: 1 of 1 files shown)


VenelinMartinov (approved):

iwahbe (approved):

iwahbe:
Looks like the test failure is an instance of https://github.com/pulumi/pulumi-gcp/issues/1387. I closed the original issue due to a missing repro, but I am re-opening.

PR #1585, opened 2024-02-12
Run one benchmark at a time

t0yv0:
Before the change, benchmarks ran in parallel. We are observing very high variance no the benchmarks, part of the problem could be resource contention for network bandwidth across benchmarks. This change will give the entire runner's worth of resources to each benchmark individually, and hopefully reduce variance at the cost of using more CI runner time. Will observe for 1 week and evaluate.

### DIFF
--- a/.github/workflows/performance_metrics_cron.yml
+++ b/.github/workflows/performance_metrics_cron.yml
@@ -90,7 +90,7 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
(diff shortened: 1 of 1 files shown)


VenelinMartinov (approved):

t0yv0:
AFAIK TestAccAwsTsHelloFargate failure should not be affected by these changes. I think I can merge, LMK if need to revert.

PR #1586, opened 2024-02-12
fix: update langchain examples

dirien:


dirien:
closes #1587

scottslowe (changes requested):
Everything looks great (love the new README files), but I don't think the Pinecone-related code should be in this PR, yes?

dirien:
@scottslowe: good catch! Removed it!

scottslowe (approved):

PR #1595, opened 2024-02-16
Use pulumi dev SDKs in CI tests

tgummerer:
Similarly how we're using the dev version of the pulumi CLI, we also want to use the dev version of the SDKs in the examples tests.  This will help us dogfood these better, and hopefully catch bugs before we release them.  We expect the dev versions to be stable and not introduce many new failures in CI here that otherwise wouldn't have been introduced.

To do this we're using the new `InstallDevReleases` flag, introduced in https://github.com/pulumi/pulumi/pull/15387

### DIFF
--- a/.github/workflows/run-tests-command.yml
+++ b/.github/workflows/run-tests-command.yml
@@ -363,6 +363,11 @@ jobs:
       with:
         version: v2.5.0
         token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Free Disk Space (Ubuntu)
+      uses: jlumbroso/free-disk-space@main
+      with:
+        tool-cache: false
+        swap-storage: false
     - name: Running ${{ matrix.clouds }}${{ matrix.languages }} Tests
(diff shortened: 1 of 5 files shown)


julienp (approved):

.github/workflows/run-tests-command.yml
```diff
@@ -363,6 +363,11 @@ jobs:
       with:
         version: v2.5.0
         token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Free Disk Space (Ubuntu)
```
tgummerer, line 366: This is necessary because the AWS Js tests were already close to running the runners out of disk space (see https://github.com/pulumi/examples/actions/runs/8060515253/job/22016676190 for example, where we did run out of disk space), and using the dev dependency seems to use just a little bit more disk space, so the failing is more consistent. Since we're already close to running the runner our of disk space this seems like a good thing to do either way.

cnunciato (approved):

PR #1596, opened 2024-02-16
fix aws-py-stepfunctions

tgummerer:
AWS no longer supports lambdas with the python3.7 runtime.  Update it to the latest runtime to get this example to work again.

I noticed this was failing in the tests currently, and the error message suggests upgrading to the latest python.


tgummerer:
Example of a failure is here: https://github.com/pulumi/examples/actions/runs/7928272420/job/21646216538

scottslowe (approved):
LGTM

PR #1600, opened 2024-02-22
set pipefail option when running tests

tgummerer:
When running the tests in CI, we're piping the output of `go test` into gotestfmt to get prettier output.  This works because `gotestfmt` will return a non-zero exit code when there are test failures.

However when there are other failures, e.g. compilation of the tests failed, `gotestfmt` will have a 0 exit code.  `go test` wouldn't, but because we're piping the results through, the shell swallows that unless we're setting the pipefail option.

Set up the pipefail option here, so tests aren't accidentally green when compilation fails.

/cc @thomas11 who noticed the tests were passing even though compilation failed in https://github.com/pulumi/examples/pull/1595


thomas11 (approved):
Thank you for the quick fix!

We should probably do a systematic review of Pulumi usages of gotestfmt without `pipefail`....

PR #1606, opened 2024-03-01
Update pulumi-trace-tool to v0.0.1

t0yv0:
The motivation is to land https://github.com/pulumi/pulumi-trace-tool/commit/ddfac4da55c8cdd3fe2bdaf01b20653eed401f1f improvement in metrics.

Looks like `go mod tidy` also updates the Go toolchain.


misc/test/go.mod
```diff
@@ -1,11 +1,13 @@
 module github.com/pulumi/examples/misc/test
 
-go 1.20
```
t0yv0, line 3: go1.20 EOL'd in Feb 6, 2024. Should be OK? Maybe merge and fix forward?

mjeffryes: looks like the tests are happy, so I see no reason not to bump to 1.21

mjeffryes (approved):
🚢

PR #1623, opened 2024-03-20
Add an example of using tsconfig-paths with Lambda layers

cnunciato:
Adds an example showing how to use `tsconfig-paths` to make locally defined Node modules work with magic Lambdas and Lambda layers. Intended to accompany the docs additions in https://github.com/pulumi/pulumi-hugo/pull/4070, as explaining exactly how to do this ended up being much too complex to include in the doc itself.


aws-ts-paths-lambda-layers/README.md
```diff
@@ -0,0 +1,28 @@
+# Using tsconfig-paths with Lambda layers
```
cnunciato, line 1: Currently working on fleshing this out.

cnunciato:
Turns out I won't need this after all.

PR #1629, opened 2024-04-09
Fix Subdomain Computation of Util function

Armadillidiid:
The `getDomainAndSubdomain` function previously incorrectly computed the subdomain for domain names with multiple subdomains. This commit fixes the computation by properly iterating over the domain parts to extract the subdomain. Additionally, it adds a trailing dot to the parent domain for canonicalization.


dirien (approved):
Clear bug fix for subdomain computation. Single file change.

PR #1631, opened 2024-04-13
chore: Replace aws-lambda with @types/aws-lambda

astuyve:
`aws-lambda` is a [poorly-named package](https://www.npmjs.com/package/aws-lambda) which is a CLI tool to deploy Lambda functions and includes heavy dependencies (the entire v2 `aws-sdk`).

It also incidentally includes types for Lambda events.

The correct package for Lambda event types is `@types/aws-lambda` as per the [docs](https://docs.aws.amazon.com/lambda/latest/dg/typescript-handler.html).

This PR adds the correct devDependency and removes the dependency on `aws-lambda`


rshade:
@astuyve I am going to merge it into this branch so that I can do a new PR so tests will run.

PR #1639, opened 2024-05-20
fix: improved region handling for `langchain` examples

dirien:
This PR improves the region handling in the `langchain` examples. Before the region was hard-coded to `eu-central-1`, which is not ideal as not everyone is using this region.

### DIFF
--- a/aws-cs-langserve/Aws.Langserve.csproj
+++ b/aws-cs-langserve/Aws.Langserve.csproj
@@ -2,7 +2,7 @@
 
 	<PropertyGroup>
 		<OutputType>Exe</OutputType>
-		<TargetFramework>net6.0</TargetFramework>
+		<TargetFramework>net8.0</TargetFramework>
 		<Nullable>enable</Nullable>
 	</PropertyGroup>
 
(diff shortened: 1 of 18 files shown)


aws-cs-langserve/Aws.Langserve.csproj
```diff
@@ -2,7 +2,7 @@
 
 	<PropertyGroup>
 		<OutputType>Exe</OutputType>
-		<TargetFramework>net6.0</TargetFramework>
+		<TargetFramework>net8.0</TargetFramework>
```
cnunciato, line 5: Curious, does this have to be .NET 8? Can it be 6? Asking because we still support 6 (and run 6 in CI), so if possible we keep to that if we can.

dirien:
```suggestion
		<TargetFramework>net6.0</TargetFramework>
```

dirien: @cnunciato ups! This is my setup but you do not need 8.0 to run it! Thanks for the feedback. I fixed it

cnunciato (approved):

PR #1641, opened 2024-05-21
update examples to use default providers

tgummerer:
This is for demonstration purposes only.


aws-py-eks/__main__.py
```diff
@@ -2,38 +2,43 @@
...
-    vpc_config=eks.ClusterVpcConfigArgs(
-        public_access_cidrs=['0.0.0.0/0'],
-        security_group_ids=[vpc.eks_security_group.id],
-        subnet_ids=vpc.subnet_ids,
-    ),
-)
+## EKS Cluster
+with pulumi.default_providers([provider]):
+    ec2_role, eks_role = iam.setup_iam()
+    eks_security_group, subnet_ids = vpc.vpc()
```
lukehoban, line 13: This is sort of the interesting bit - and also where "global" approach would run into some problems - I assume the dependent modules would all be loaded and run before we could "install" the default provider?

That said - this does lead to a decent amount more code refactoring to get to something that works?

aws-py-eks/iam.py
```diff
@@ -1,72 +1,73 @@
 from pulumi_aws import config, iam
 import json
 
-## EKS Cluster Role
+def setup_iam():
```
lukehoban, line 4: This is sort of introducing another new kind of "nesting" - beyond:
1. Components
2. Python modules

I suspect in practice we would want to encourage refactoring into components in these cases?

aws-ts-eks/index.ts
```diff
@@ -1,20 +1,30 @@
...
-// Create the EKS cluster itself and a deployment of the Kubernetes dashboard.
-const cluster = new eks.Cluster("cluster", {
-    vpcId: vpc.vpcId,
-    subnetIds: vpc.publicSubnetIds,
-    instanceType: "t2.medium",
-    desiredCapacity: 2,
-    minSize: 1,
-    maxSize: 2,
+let cluster: eks.Cluster | undefined;
+pulumi.withDefaultProviders([provider, providerVersioned], () => {
```
lukehoban, line 12: What does this do?  How does the code within here consume these two different AWS providers?

PR #1642, opened 2024-05-21
More Deploy button fixes

cnunciato:
Handful of quick changes to make the repo more compatible with imported examples guides:

* Uses the updated buttons from get.pulumi.com instead of www (I added a light button there as well)
* Removes some duplicate buttons (e.g., to [fix things like this](https://www.pulumi.com/registry/packages/aws/how-to-guides/aws-ts-pulumi-miniflux/))
* Adds light-mode buttons where there aren't any for consistency


sean1588 (approved):
lgtm, and good catch with the duplicate buttons.

PR #1643, opened 2024-06-07
Add examples for the Command provider

thomas11:
- **Command: add simple TS example for a local command**
- **Command: add simple YAML example for a local command**

Draft PR because at least one more example, for the Copy resource, is coming.


thomas11:
We've discussed this and would rather have the examples repo focus on larger examples that show E2E scenarios.

PR #1644, opened 2024-06-08
Update testing-unit-ts dependencies

justinvp:
I tried running the example locally with `npx mocha -r ts-node/register ec2tests.ts` and ran into an error. However, after upgrading the dependencies to the latest versions, everything works as expected.

Updated the CI workflows to use the same `npx mocha ...` command we mention in https://github.com/pulumi/examples/blob/master/testing-unit-ts/mocha/README.md, and updated those to run on the current active LTS version of Node (20).

Fixes https://github.com/pulumi/examples/issues/1645

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -613,9 +613,9 @@ jobs:
     runs-on: ${{ matrix.platform }}
     steps:
     - name: Checkout Repo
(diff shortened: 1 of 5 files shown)


cnunciato (approved):
Thanks @justinvp!

PR #1646, opened 2024-06-10
Update RDS instance types to db.t3.micro

cnunciato:
Noticed the Wordpress + Fargate example was failing because `db.t2.micro` [is deprecated](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html#Concepts.DBInstanceClass.Support). This fixes things up.


interurban (approved):
thx !

PR #1654, opened 2024-06-27
Update examples to Command provider 1.0

thomas11:
This PR updates all examples that use the Command provider to use ~~1.0-beta.1~~ 1.0.0.

Note on Java: there are no Command-related changes but the example doesn't compile with the `source` version set to less than 15 so I fixed that.


classic-azure-ts-vm-provisioners/index.ts
```diff
@@ -139,11 +139,12 @@ const connection: command.types.input.remote.ConnectionArgs = {
...
 
 const changeToken = getFileHash("myapp.conf");
+
 // Copy a config file to our server.
-const cpConfig = new command.remote.CopyFile("config", {
+const cpConfig = new command.remote.CopyToRemote("config", {
     triggers: [changeToken],
     connection,
-    localPath: "myapp.conf",
+    source: new pulumi.asset.FileArchive("myapp.conf"),
```
lukehoban, line 147: Intentional that this uses FileArchive vs FileAsset?

classic-azure-ts-vm-provisioners/package.json
```diff
@@ -3,8 +3,8 @@
     "version": "0.1.0",
     "main": "index.js",
     "dependencies": {
-        "@pulumi/azure": "^4.0.0",
-        "@pulumi/command": "^0.0.3",
+        "@pulumi/azure": "^5.0.0",
+        "@pulumi/command": "1.0.0-beta.1",
```
lukehoban, line 7: Looks like this example may not be covered by tests in misc/tests?  Is it worth adding it there to make sure these changes work and that we don’t regress in the future?

lukehoban (approved):

classic-azure-ts-vm-provisioners/package.json
thomas11: Filed #1658 so we don't lose track of it and I can merge this PR.

PR #1656, opened 2024-07-02
Use dictionary literals, poetry & pyright in aws-py-fargate

julienp:
Use the latest python features from https://github.com/pulumi/docs/pull/12123 in the aws-py-fargate example.

This could also go into a separate example instead of changing the existing one, not sure what we prefer here?


tgummerer (approved):
Seems reasonable to me to update an existing example with this.

PR #1663, opened 2024-07-05
Do some cleanup

### LINKED ISSUE #1661: Refactor run-tests-command.yml
This is a very large file that has a ton of duplication and is much harder to understand than it needs to be. We should clean it up and remove what we can to make it more maintainable.

cnunciato:
Fixes #1661.


.github/workflows/test-examples.yml
```diff
@@ -0,0 +1,224 @@
...
+  lint-ts:
+    runs-on: pulumi-ubuntu-8core
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up the environment
+        uses: ./.github/actions/setup
```
sean1588, line 23: I like this!

sean1588 (approved):
LGTM!

PR #1679, opened 2024-09-06
Test v8 of pulumi-gcp provider

guineveresaenger:
Test success of pulumi GCP v8 on alpha cut


guineveresaenger:
Relevant tests are green. QED; closing.

PR #1687, opened 2024-09-17
Update aws-fs-s3-folder to BucketV2

t0yv0:
Re: pulumi/home#3631


mjeffryes (approved):
Seems questionable whether we really want to keep maintaining F# examples.

t0yv0:
It's the best language out there. One day it will draw contributors I'm certain.

PR #1692, opened 2024-09-18
Update aws-py-static-website to BucketV2

t0yv0:
Re: pulumi/home#3631

### DIFF
--- a/aws-py-static-website/__main__.py
+++ b/aws-py-static-website/__main__.py
@@ -10,6 +10,33 @@ import pulumi_aws.config
 import pulumi_aws.route53
 import pulumi_aws.s3
 
+
+def setup_acl(bucket_name, bucket, acl):
+    bucket_ownership_controls = pulumi_aws.s3.BucketOwnershipControls(
+        bucket_name,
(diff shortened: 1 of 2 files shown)


mjeffryes (approved):

aws-py-static-website/__main__.py
```diff
@@ -10,6 +10,33 @@
...
+
+def setup_acl(bucket_name, bucket, acl):
+    bucket_ownership_controls = pulumi_aws.s3.BucketOwnershipControls(
+        bucket_name,
+        bucket=bucket.bucket,
+        rule={
+            "object_ownership": "BucketOwnerPreferred",
+        })
+
+    public_access_block = pulumi_aws.s3.BucketPublicAccessBlock(
```
t0yv0, line 22: I'm having second thoughts on whether this is necessary, will double check with @flostadler  and @corymhall  on Monday.

t0yv0: OK this is necessary to be able to serve the website off the bucket on the public network. Confirmed with team. AWS is making this harder intentionally and this is reflected in the TF mapping.

PR #1699, opened 2024-09-19
Update aws-ts-static-website to BucketV2

t0yv0:
Re: pulumi/home#3631

### DIFF
--- a/aws-ts-static-website/index.ts
+++ b/aws-ts-static-website/index.ts
@@ -27,6 +27,7 @@ const config = {
 
 // contentBucket is the S3 bucket that the website's contents will be stored in.
 const contentBucket = new aws.s3.BucketV2("contentBucket", {bucket: config.targetDomain});
+configureACL("contentBucket", contentBucket, "public-read");
 
 const contentBucketWebsite = new aws.s3.BucketWebsiteConfigurationV2("contentBucketWebsite", {
     bucket: contentBucket.bucket,
@@ -61,8 +62,6 @@ crawlDirectory(
             relativeFilePath,
             {
                 key: relativeFilePath,
-
-                acl: "public-read",
                 bucket: contentBucket.bucket,
                 contentType: mime.getType(filePath) || undefined,
                 source: new pulumi.asset.FileAsset(filePath),
--- a/aws-ts-static-website/index.ts
+++ b/aws-ts-static-website/index.ts
@@ -26,8 +26,7 @@ const config = {
 };
 
 // contentBucket is the S3 bucket that the website's contents will be stored in.
-const contentBucket = new aws.s3.BucketV2("contentBucket", {bucket: config.targetDomain});
-configureACL("contentBucket", contentBucket, "public-read");
+const contentBucket = new aws.s3.BucketV2(`${config.targetDomain}-content`);
 
 const contentBucketWebsite = new aws.s3.BucketWebsiteConfigurationV2("contentBucketWebsite", {
     bucket: contentBucket.bucket,
@@ -72,8 +71,7 @@ crawlDirectory(
     });
 
 // logsBucket is an S3 bucket that will contain the CDN's request logs.
-const logsBucket = new aws.s3.BucketV2("requestLogs", { bucket: `${config.targetDomain}-logs` });
-
+const logsBucket = new aws.s3.BucketV2(`${config.targetDomain}-logs`);
 configureACL("requestLogs", logsBucket, "private");
 
 const tenMinutes = 60 * 10;
--- a/aws-ts-static-website/index.ts
+++ b/aws-ts-static-website/index.ts
@@ -5,7 +5,7 @@ import * as pulumi from "@pulumi/pulumi";
 import * as fs from "fs";
 import * as mime from "mime";
 import * as path from "path";
-
+import { configureACL } from "./acl";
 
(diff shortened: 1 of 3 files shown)


mjeffryes (approved):

aws-ts-static-website/index.ts
```diff
@@ -76,11 +73,9 @@ crawlDirectory(
...
 
 // logsBucket is an S3 bucket that will contain the CDN's request logs.
-const logsBucket = new aws.s3.Bucket("requestLogs",
-    {
-        bucket: `${config.targetDomain}-logs`,
-        acl: "private",
-    });
+const logsBucket = new aws.s3.BucketV2("requestLogs", { bucket: `${config.targetDomain}-logs` });
+
+configureACL("requestLogs", logsBucket, "private");
```
corymhall, line 78: We don't need to configure the ACL on this bucket since it is a private bucket. By default ACLs are disabled and block public access is turned on for new buckets.

corymhall (changes requested):

aws-ts-static-website/index.ts
```diff
@@ -66,7 +63,7 @@ crawlDirectory(
                 key: relativeFilePath,
 
                 acl: "public-read",
```
corymhall, line 65:
```suggestion
```

With the new default settings ACLs are disabled so we can't/shouldn't set acls on the object itself.

corymhall (changes requested):

aws-ts-static-website/index.ts
```diff
@@ -26,16 +26,13 @@ const config = {
...
-    {
-        bucket: config.targetDomain,
-        // Configure S3 to serve bucket contents as a website. This way S3 will automatically convert
-        // requests for "foo/" to "foo/index.html".
-        website: {
-            indexDocument: "index.html",
-            errorDocument: "404.html",
-        },
-    });
+const contentBucket = new aws.s3.BucketV2("contentBucket", {bucket: config.targetDomain});
```
flostadler, line 29: Don't we have to make this bucket publicly readable?

aws-ts-static-website/index.ts
```diff
@@ -66,7 +63,7 @@ crawlDirectory(
                 key: relativeFilePath,
```
flostadler, line 63: Should this use BucketObjectV2?

aws-ts-static-website/index.ts
t0yv0: Results in:

```
  aws:cloudfront:Distribution (cdn):
    error:   sdk-v2/provider2.go:472: sdk.helper_schema: creating CloudFront Distribution: operation error CloudFront: CreateDistributionWithTags, https response error StatusCode: 400, RequestID: 598a5505-a016-49a6-a6d3-037c73122459, InvalidArgument: The S3 bucket that you specified for CloudFront logs does not enable ACL access: aws-ts-static-website.pulumi-demos.net-logs.s3.amazonaws.com: provider=aws@6.52.0

```

t0yv0: Adding it back gives

```
    error:   sdk-v2/provider2.go:472: sdk.helper_schema: putting S3 Object (404.html) ACL: operation error S3: PutObjectAcl, https response error StatusCode: 403, RequestID: 0AK7AYSN54Z5V79V, HostID: eqG88AMTZddh2iV3y0yTafd84K0rZtgEMETB/Q1Kim6S6YF2Au0Rhp7RoRHVmuwt+QdWIeM/6xw=, api error AccessDenied: User: arn:aws:sts::616138583583:assumed-role/AWSReservedSSO_AdministratorAccess_224e23fb315b65cb/anton@pulumi.com is not authorized to perform: s3:PutObjectAcl on resource: "arn:aws:s3:::aws-ts-static-website.pulumi-demos.net/404.html" because public access control lists (ACLs) are blocked by the BlockPublicAcls block public access setting.: provider=aws@6.52.0

```

aws-ts-static-website/index.ts
t0yv0: I think that's right. The example doesn't provision otherwise. Checking.

aws-ts-static-website/index.ts
t0yv0: Similar to other comments, trying to avoid moving this to avoid the move back later on in aws V7. BucketObject and BucketObjectV2 are really really close.

aws-ts-static-website/index.ts
t0yv0: OK, I think I got this cleaned out.

t0yv0:
PTAL.

aws-ts-static-website/index.ts
corymhall: I don't think we do since we are configuring public access through a CloudFront distribution with accesses the bucket via an OIC and bucket policies.

corymhall (dismissed):

aws-ts-static-website/index.ts
t0yv0: Your intuition is right. I mixed up some test results. It works without it, therefore I will remove.

t0yv0: I think sadly ACL is still needed for the logs bucket. Just not the content bucket.

t0yv0: Let me update Python to match.

PR #1704, opened 2024-09-19
Update secrets-provider to BucketV2

t0yv0:
Re: pulumi/home#3631

### DIFF
--- a/secrets-provider/aws/README.md
+++ b/secrets-provider/aws/README.md
@@ -64,7 +64,7 @@ pulumi up --yes
 Previewing update (aws-kms):
      Type                    Name                    Plan
  +   pulumi:pulumi:Stack     pulumi-aws-kms-aws-kms  create
- +   ├─ aws:s3:Bucket        bucket                  create
+ +   ├─ aws:s3:BucketV2      bucket                  create
  +   └─ aws:s3:BucketObject  secret                  create
 
 Resources:
@@ -73,7 +73,7 @@ Resources:
 Updating (aws-kms):
      Type                    Name                    Status
  +   pulumi:pulumi:Stack     pulumi-aws-kms-aws-kms  created
- +   ├─ aws:s3:Bucket        bucket                  created
+ +   ├─ aws:s3:BucketV2      bucket                  created
  +   └─ aws:s3:BucketObject  secret                  created
 
 Outputs:
@@ -100,9 +100,3 @@ pulumi up --yes
 error: getting secrets manager: secrets (code=Unknown): InvalidSignatureException: The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.
 	status code: 400, request id: 35ff51c6-ef88-4c06-9146-361231b8fd4a
 ```
-
-
-
-
-
-
--- a/secrets-provider/aws/index.ts
+++ b/secrets-provider/aws/index.ts
@@ -10,11 +10,35 @@ const config = new pulumi.Config();
 const bucketName = config.require('bucketName');
 const secretValue = config.requireSecret('secretValue');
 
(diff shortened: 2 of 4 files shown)


secrets-provider/aws/index.ts
```diff
@@ -10,12 +10,36 @@ const config = new pulumi.Config();
 const bucketName = config.require('bucketName');
 const secretValue = config.requireSecret('secretValue');
 
+export function configureACL(bucketName: string, bucket: aws.s3.BucketV2, acl: string): aws.s3.BucketAclV2 {
```
corymhall, line 13: For this example I don't think we need to do anything with the ACL. By default ACLs are disabled and block public access is turned on for new buckets. Since this looks to be a normal bucket (not a public website bucket) I think the default is what we want.

secrets-provider/vault/index.ts
```diff
@@ -10,12 +10,36 @@ const config = new pulumi.Config();
 const bucketName = config.require('bucketName');
 const secretValue = config.requireSecret('secretValue');
 
+function configureACL(bucketName: string, bucket: aws.s3.BucketV2, acl: string): aws.s3.BucketAclV2 {
```
corymhall, line 13: Same as the other comment.

corymhall (dismissed):

secrets-provider/aws/README.md
```diff
@@ -64,7 +64,7 @@ pulumi up --yes
 Previewing update (aws-kms):
      Type                    Name                    Plan
  +   pulumi:pulumi:Stack     pulumi-aws-kms-aws-kms  create
- +   ├─ aws:s3:Bucket        bucket                  create
+ +   ├─ aws:s3:BucketV2      bucket                  create
  +   └─ aws:s3:BucketObject  secret                  create
```
flostadler, line 68: Shouldn't this also update to `BucketObjectV2` or is that resource still fine?

secrets-provider/aws/index.ts
```diff
@@ -10,12 +10,36 @@ const config = new pulumi.Config();
...
+    const ownership = new aws.s3.BucketOwnershipControls(bucketName, {
+        bucket: bucket.bucket,
+        rule: {
+            objectOwnership: "BucketOwnerPreferred",
```
flostadler, line 17: `BucketOwnerEnforced` (All ACLs disabled), the default setting for S3 buckets, is preferred if no public access is needed.

secrets-provider/aws/index.ts
```diff
@@ -10,12 +10,36 @@ const config = new pulumi.Config();
...
+        }
+    });
+    const publicAccessBlock = new aws.s3.BucketPublicAccessBlock(bucketName, {
+        bucket: bucket.bucket,
+        blockPublicAcls: false,
+        blockPublicPolicy: false,
+        ignorePublicAcls: false,
+        restrictPublicBuckets: false,
+    });
+    const bucketACL = new aws.s3.BucketAclV2(bucketName, {
```
flostadler, line 27: This isn't in scope for this PR, but it feels a bit weird to upload a "secret value" to an S3 bucket that doesn't have a custom KMS key, does not enforce TLS and doesn't have versioning enabled

secrets-provider/aws/README.md
t0yv0: I've avoided rewriting BucketObject into BucketObjectV2 intentionally. These currently share implementation but are distinct TF resources with identical schema. I'm hoping we can silently cut over to BucketObjectV2 implementation in pulumi-aws V7.

flostadler: Fair enough! I left similar comments on 1-2 other PRs, feel free to just resolve them

corymhall (approved):

PR #1708, opened 2024-09-20
Remove aws-native examples

mjeffryes:
Removes most examples using aws-native; especially examples that just duplicate an example using the pulumi-aws provider.

I decided to retain `aws-yaml-static-website` for now since it seems to be somewhat unique.

fixes: https://github.com/pulumi/pulumi-aws-native/issues/1702


aws-ts-synthetics-canary/package.json
```diff
@@ -5,7 +5,6 @@
     },
     "dependencies": {
         "@pulumi/aws": "^6.0.2",
-        "@pulumi/aws-native": "^0.15.0",
```
mjeffryes, line 8: there's not any use of this provider in the actual code

corymhall (approved):

PR #1710, opened 2024-10-02
Azure Classic v6 updates

danielrbradley:
- Upgrade dependencies
- Add subscriptionId config to READMEs (and Pulumi.yaml where template exists).
- Reformat markdown:
  - Remove "$" prefixes to make the bash valid.
  - Set mixed bash inputs and console outputs to use "console".

Closes https://github.com/pulumi/pulumi-azure/issues/2336


thomas11 (approved):

thoward:
@danielrbradley hi, i'm following up on a 404 from our documentation link checker for a link to the [`classic-azure-ts-arm-template`](https://github.com/pulumi/examples/tree/master/classic-azure-ts-arm-template) directory. It seems to have been deleted in this PR, and I'm not sure if that was intentional or not from the context here.

thoward:
Update, I see from this commit that it *was* intentional: https://github.com/pulumi/examples/pull/1710/commits/c2f6d8a604f711c57c987cb8a9c5d1b0433b5123 ... this was linked from a 2019-era blog post: https://www.pulumi.com/blog/2019-year-at-a-glance/ ... and also, we have an article on it that links to this code here: https://www.pulumi.com/registry/packages/azure/how-to-guides/classic-azure-ts-arm-template/ 

Should we remove all references to this directory or restore the directory and update the text there to label it as deprecated?

thoward:
There are similar dead link issues for all the other example directories removed in that commit (e.g. `classic-azure-ts-aks-with-diagnostics`, `classic-azure-ts-dynamicresource`, `classic-azure-ts-aks-helm`, `classic-azure-ts-appservice-springboot`, etc). Each one has multiple blogs/docs with dead links to this code now.

danielrbradley:
Thanks for spotting this. These aren't maintained and so should be removed to avoid confusion for users browsing the examples, but I understand broken links are not good. 

I think what would make sense for these very old articles is to update them with permalinks to the last maintained version:
- https://github.com/pulumi/examples/tree/846811de2c7faa4694454c64edc9bbcdb31d533e/classic-azure-ts-aks-with-diagnostics
- https://github.com/pulumi/examples/tree/846811de2c7faa4694454c64edc9bbcdb31d533e/classic-azure-ts-dynamicresource
- https://github.com/pulumi/examples/tree/846811de2c7faa4694454c64edc9bbcdb31d533e/classic-azure-ts-aks-helm
- https://github.com/pulumi/examples/tree/846811de2c7faa4694454c64edc9bbcdb31d533e/classic-azure-ts-appservice-springboot

Does that work for you @thoward?

Not sure how impactful or common of an issue we consider this .. I suppose we _could_ theorecically proactively check for referenceing code before we modify or remove any in the future.

thoward:
Yes! linking to the most recent commit hash will work in this case.

PR #1711, opened 2024-10-03
Revert "Update aws-ts-s3-lambda-copyzip to BucketV2"

t0yv0:
Reverts pulumi/examples#1696

Apologies, somehow merged over a failing test. This is not ready to go yet because of https://github.com/pulumi/pulumi-aws/issues/4475


t0yv0:
That's a good point, so like expand and inline the onObject*? It's a bit laborious but may be worth it. I can add this to migration guide as well I guess. This comes up a lot also in the registry content.

t0yv0:
The other option here is to maybe just add onObject* to pulumi-aws BucketV2. It might not be very difficult.

PR #1712, opened 2024-10-05
Fix OIDC TS example.

jkodroff:
Update the OIDC ESC setup example to account for ESC projects and the newer OIDC subject claims format.


komalali (approved):

pierskarsenbarg (approved):
LGTM

PR #1719, opened 2024-11-01
Fix AWS OIDC TS example

jkodroff:
It's not possible to have a single stack which handles both the use case of creating a new OIDC provider and also adding an audience if none exists. This change removes the conditional creation of an OIDC provider and assumes none exists.


pierskarsenbarg (approved):

pierskarsenbarg (changes requested):
Couple of linting errors to fix

pierskarsenbarg (approved):

thoward:
Looks good. Thanks for this! Would love to see an update to the README that explains this limitation.

thoward:
Merging for @jkodroff per request.

PR #1721, opened 2024-11-05
New example: testing-unit-go-components

EronWright:
Closes https://github.com/pulumi/pulumi-kubernetes/issues/3183


blampe (approved):
This is great! Quite involved but very glad you were able to figure it out.

PR #1738, opened 2024-12-11
Add yaml example that does math and uses pulumi-std functions

MitchellGerdisch:
In addition to programming languages such as Python, Go, Typescript, C#, Pulumi also supports YAML.
However, YAML is limited when it comes to constructs such as loops and conditionals and it is not always obvious how one can do math in a Pulumi YAML program.
Similarly, using the `pulumi-std` library can be a little tricky.

This project is meant to provide some YAML program examples for these use-cases.


automagic (approved):

mjeffryes (approved):
Thanks! This looks like it will be helpful reference for users.

PR #1761, opened 2025-01-09
Swapping out GKE for MiniKube

rshade:
This PR Swaps out GKE for minikube, it saves us on money and should also help with speeding up kubernetes runs. Down to 13m.

### DIFF
--- a/.github/workflows/test-examples.yml
+++ b/.github/workflows/test-examples.yml
@@ -213,13 +213,19 @@ jobs:
           aws-role-to-assume: ${{ secrets.AWS_CI_ROLE_ARN }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Setup KinD cluster
+        uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
+        with:
+          cluster_name: kind-integration-tests-examples
+          node_image: kindest/node:v1.29.2
+
       - name: Run tests
         run: |
-          ./misc/scripts/create-ci-cluster.sh "${{ env.INFRA_STACK_NAME }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-examples.yml
```diff
@@ -213,13 +213,40 @@ jobs:
           aws-role-to-assume: ${{ secrets.AWS_CI_ROLE_ARN }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Set up Minikube
+        uses: manusa/actions-setup-minikube@v2.13.0
+        with:
+          minikube version: 'v1.34.0'
+          kubernetes version: 'v1.29'
```
rquitales, line 220: Is there a reason to use v1.29? Could we drop this and use the default k8s version used by the configured minikube version?

.github/workflows/test-examples.yml
```diff
@@ -213,13 +213,40 @@ jobs:
...
+          driver: docker
+          start args: --cpus 2 --memory 4096 --addons metallb --wait all
+
+      - name: Start Minikube
```
rquitales, line 224: We don't need this step since the previous step already starts a minikube cluster with metallb enabled.

.github/workflows/test-examples.yml
rshade: I just pinned it to what kind was using, but yes I am all for rolling forward

rshade: Looks like `kubernetes version` is required, so I pinned it to the version here: https://github.com/manusa/actions-setup-minikube?tab=readme-ov-file#basic

rquitales (approved):

rshade:
Awesome @rquitales thank you for all the help and guidance on this PR.

PR #1762, opened 2025-01-09
Reformatting everything with black and adding a test for it

rshade:
This PR reformats all the python files with black, so they have a uniform formatting. It adds a make command and test.


smithrobs (approved):
I like consistency 😄 Thanks for adding `black.toml`; LGTM

smithrobs (changes requested):
hm why did the one check fail?

PR #1764, opened 2025-01-09
Add renovate.json to help updates process cleanly

### LINKED ISSUE #1757: renovate: don't group dependencies
We have so many dependencies that upgrading them altogether is increasingly unlikely to succeed.

We should define a `renovate.json5` file in this repo and configure it to no longer group first-party dependencies.

We should also probably modify to run on a schedule, maybe daily, because this repo takes quite a while to process.

rshade:
Fixes #1757

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -12,6 +12,13 @@
         'crd2pulumi-crontabs/kubernetes-go-crontabs/**',
       ],
       enabled: false,
-    },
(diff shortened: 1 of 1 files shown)


rshade:
@danielrbradley with @blampe out would you mind reviewing this?

mjeffryes (approved):

PR #1768, opened 2025-01-09
Update dotnet examples to net8.0

smithrobs:
Same rationale as https://github.com/pulumi/templates/issues/690 - as dotnet 6.0 will soon be EOL, move examples to dotnet 8.0 LTS


rshade (approved):
LGTM

PR #1787, opened 2025-01-14
Remove use of deprecated `getId` methods in Java examples

lunaris:
The `getId()` method on resources in Java is `@Deprecated`, and has been for some time. This commit removes uses of this method in favour of the recommended `id()`, which provides exactly the same behaviour and is consistent with other methods in the Java SDK/generated SDKs.


rshade (approved):
LGTM

PR #1803, opened 2025-01-17
[DRAFT] revert updates from 68a196b0750e35471e51fe7f9d223709a2bff7b7 in the test dir

brandonpollack23:


rshade:
@brandonpollack23 do you mean automated or manual. We could probably hack out a script that ran it a bunch of times, it wouldn't be cool and automated, but it might get us what we wanted, do we also need to increment up pulumi to find when the policy packs started erroring.

rshade:
Also, do we want to just revert aws, and testify in this PR to see if its pulumi or one of those, then next pr, do pulumi if its pulumi

dirien:
Closing: draft PR that's been open 14 months, appears to be an experimental benchmark that was never completed.

PR #1831, opened 2025-02-04
Update deploy button svg references in example readmes

sean1588:
The deploy button svg hosted at [get.pulumi.com/new/button.svg](get.pulumi.com/new/button.svg) has been causing problems lately and has been very flaky causing the button image to fail to render. We have this asset hosted on the www.pulumi.com website at https://www.pulumi.com/images/deploy-with-pulumi/dark.svg, so update to use that instead.


thoward (approved):
Looks good!

meagancojocar (approved):
LGTM (I just reviewed a sample)

PR #1844, opened 2025-02-12
Chore: Removing devcontainer as it is archived

rshade:
Dependabot fails updates because of devcontainer it looks like: https://github.com/pulumi/examples/actions/runs/13272735598/job/37055745754


zbuchheit (approved):
lgtm!

PR #2026, opened 2025-02-21
Use basic runners for  test-examples.yml

mjeffryes:
Examples testing can consume a lot of GHA runner minutes; this is a public repo, so we might as well just limit ourselves to the free runners.

### DIFF
--- a/.github/workflows/test-examples.yml
+++ b/.github/workflows/test-examples.yml
@@ -168,6 +168,14 @@ jobs:
       contents: read
 
     steps:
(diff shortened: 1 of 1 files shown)


smithrobs (approved):
Makes sense to me!

smithrobs:
AwsTs tests seem to be consistently failing due to storage limits...

mjeffryes:
> AwsTs tests seem to be consistently failing due to storage limits...

I've added a step we use in the providers builds to free up disk space 🤞

PR #2168, opened 2025-04-16
fix dependencies for azure-java-appservice-sql example

tgummerer:
This example doesn't work with azure-native 3.x.  However our dependencies currently *include* 3.0, instead of excluding it, which was probably what was intended here.

This example also should keep working with pulumi-java `1.x`. Currently we allow `1.0.0`, but no further versions, which is also incorrect.  Fix that, to include 1.x, but exclude 2.x.

This fixes the test failure in pulumi-java (see the failing job in https://github.com/pulumi/pulumi-java/pull/1767)


mikhailshilkov (approved):
Nice catch!

PR #2181, opened 2025-05-23
Adding in claude worker

rshade:


dirien:
Closing this one — the workflow has an `if` condition checking for `pulumi/pulumi-afp`, so it would never fire in this repo. The repo also has AGENTS.md and CLAUDE.md set up now, which covers the AI-assisted workflow side of things.

PR #2183, opened 2025-06-11
Remove Python versions of OIDC/Pulumi Cloud

jkodroff:
- The Python versions don't have the right audience claim.
- The TS versions do, and we don't gain much from maintaining the codebase in multiple languages.
- Add CODEOWNERS with ownership for AWS and GCP OIDC setup.


jkodroff:
Need to check whether there are links to this codebase from the docs.

PR #2205, opened 2025-11-20
enable journaling in examples CI

tgummerer:
Now that all tests pass in pulumi/pulumi CI when journaling is enabled (https://github.com/pulumi/pulumi/pull/20990) also enable it in this repo.


tgummerer:
The test failures here are unrelated to journaling. Trying to fix them in https://github.com/pulumi/examples/pull/2207 and https://github.com/pulumi/examples/pull/2206.

Frassle (approved):

PR #2207, opened 2025-11-20
fix F# testing

tgummerer:
Try to fix the error we get in CI. No idea if this actually works, but CI will tell us.


tgummerer:
Hrm this isn't helping, so does that mean `null` is not of type `obj` then?  @Frassle @Zaid-Ajaj I'm pretty sure one of you is going to know what `Type constraint mismatch. The type �    ''a'    �is not compatible with type�    'obj'` means :)

Frassle:
This is odd, it runs fine on my machine with dotnet 8 and 9.

Frassle:
Ah ok its a dotnet 10 change, looks like you can just remove the cast and return null directly.

PR #2271, opened 2026-01-26
Revert "Update dependency mime to v4 (#2007)"

justinvp:
This reverts commit 2121bafe51746d6c1d4cceabdd9f434f391822f7.

Fixes #2270

PR #2272, opened 2026-01-26
Fix examples broken by `mime` v3 -> v4 upgrade

### LINKED ISSUE #2270: aws-ts-s3-folder is broken, causing CLI perf tests to fail
`aws-ts-s3-folder` was broken by https://github.com/pulumi/examples/pull/2007 (likely all the examples that were updated are now broken), causing CLI perf tests to fail.

https://github.com/pulumi/examples/pull/2007 updated the `mime` dependency from v3 to v4, but v4 requires ESM modules and these examples aren't configured for ESM modules. https://github.com/broofa/mime?tab=readme-ov-file#mime

justinvp:
This is an alternative to [reverting the change](https://github.com/pulumi/examples/pull/2271).

#2007 upgraded the `mime` dependency from `v3` to `v4`. This change broke these examples, as `mime` `v4` requires the use of ESM modules. In particular, this broke the `aws-ts-s3-folder` example, which is used by our CLI performance tests.

This change fixes the examples to use ESM modules. I've confirmed the fix for `aws-ts-s3-folder`. The remaining examples were updated in the same way by Claude.

Fixes #2270

### DIFF
--- a/aws-ts-eks-gpu-dra/index.ts
+++ b/aws-ts-eks-gpu-dra/index.ts
@@ -1,8 +1,10 @@
-import * as pulumi from "@pulumi/pulumi";
+// Copyright 2016-2025, Pulumi Corporation.  All rights reserved.
+
 import * as aws from "@pulumi/aws";
 import * as awsx from "@pulumi/awsx";
 import * as eks from "@pulumi/eks";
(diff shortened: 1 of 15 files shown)


justinvp:
LGTM

PR #2969, opened 2026-08-13
Add GOOGLE_PROJECT to OIDC setup for GCloud

jkodroff:
The GOOGLE_PROJECT env var will be reported missing if a user attempts to do a GCloud account scan, so we need to add it here in addition to the pulumiConfig output.


smithrobs (approved):
LGTM

PR #2972, opened 2026-08-13
Run only the examples a PR touched, keep a nightly full sweep

### LINKED ISSUE #666: Review what tests are currently running in CI
There are a number of tests that are not running in CI but that should be - we need to have another round of ensuring as many of these tests run as possible

### LINKED ISSUE #1245: Run all examples touched by a pr as ci tests
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

It's a heads up.

1) As reported in several places (e.g. https://github.com/pulumi/examples/issues/395 or https://github.com/pulumi/examples/issues/666 or https://github.com/pulumi/examples/issues/1057) not all examples are tested by github actions. Some documentation examples break and are reported by end-users, but the number of unreported, broken examples is unknown and circulates as a word of mouth (e.g. https://github.com/pulumi/pulumi-azure-native/issues/1495#issuecomment-1171523665). Sometimes broken examples are being disabled and may not come back soon (e.g. https://github.com/pulumi/pulumi-azure-native/issues/898 or https://github.com/pulumi/pulumi-azure-native/pull/1599 or https://github.com/pulumi/pulumi-azure-native/issues/304 or https://github.com/pulumi/examples/issues/1188).

2) The newest successful status of "Run Examples Tests From PR" github action workflow appears 25 days ago and the last successful "Smoke Test Specific Version of CLI" two months ago.

Screenshot from 2022-06-30 21-17-59
Screenshot from 2022-06-30 21-22-38

3) Pull requests with a failed status of "Run Examples Tests From PR" are merged. They may fail on unrelated tests, but don't actually test the focus feature of the PR (let's take for example https://github.com/pulumi/examples/pull/1234).

4) The tests output a red "Error" when the run appears successful (folllowing the example of https://github.com/pulumi/examples/pull/1234 above)

Screenshot from 2022-06-30 21-42-44

5) The native providers have limited coverage of resource tests. The usage of the native examples is not clear: on one hand they are tested, but on the other hand not recommended for end-users https://github.com/pulumi/pulumi-azure-native/issues/1043#issuecomment-899854236.

### Affected area/feature

https://github.com/pulumi/examples, but also other repos that contain examples.

jkodroff:
Fixes #1245
Fixes #666

## Problem

Every PR to `master` runs the **entire** integration matrix — 16 `<Cloud><Lang>` jobs plus Kubernetes, each doing a real `pulumi up` → refresh → `destroy` against staging — regardless of what the PR changed. The recent CONTRIBUTING.md-only PR (#2970) took ~20 minutes and provisioned cloud resources across five providers to validate a documentation edit.

Beyond the cost, it costs *signal*. Because the full matrix is flaky in aggregate, a red X says nothing about the PR itself, which is how PRs end up merged with failing checks — one of the specific complaints in #1245.

## Approach

Split the two concerns into two workflows:

- **`pull-request.yml`** (new) runs the same real integration tests, scoped to the examples the PR touched.
- **`test-examples.yml`** keeps the nightly cron, `workflow_dispatch`, and the fork-PR `repository_dispatch` path, and still runs everything so untouched examples don't silently rot. **Only its `pull_request` trigger is removed** — the recurring sweep is otherwise untouched and carries no risk from this change.

Narrowing happens in two layers, because either alone leaks:

1. **`scripts/changed-examples.sh`** turns the diff into a dynamic `fromJson` matrix, so untouched cloud/language combos never start a job.
2. **`PULUMI_CHANGED_PATHS`** narrows *within* a job. This one is essential: `make specific_test_set TestSet=AwsTs` filters on an **unanchored** `TestAccAwsTs` prefix, so without it a one-example change still deploys all ~30 AwsTs examples.

`helpers.ProgramTest` reads each test's own `Dir` rather than consulting a directory→test-name table, so there's nothing to keep in sync and it works unchanged for the table-driven Azure (`definitions/`) and Kubernetes tests.

## Notes for reviewers

**New tests must call `helpers.ProgramTest`, not `integration.ProgramTest`** — otherwise they opt out of scoping and deploy on every PR. Documented in `misc/test/AGENTS.md`. `performance_test.go` is deliberately exempt.

**Matching is on path branches, not example names.** The subprojects of a multi-project example stay independent: a change under `kubernetes-go-guestbook/simple` does not deploy `components`, while a change to a file directly under `kubernetes-go-guestbook/` covers both.

**Empty means unscoped.** Actions sets an env var to `""` rather than leaving it unset, so on a full sweep `PULUMI_CHANGED_PATHS=""` would otherwise mean "skip everything" — quietly turning the nightly sweep into a no-op. There's a regression test pinning this.

**Shared-path escape hatch.** Changes to `misc/test/**`, `.github/**`, `scripts/**`, `Makefile`, or root `package.json` fall back to the full sweep, since they can affect any example. **That includes this PR**, so expect a full 16-way run here — which doubles as the end-to-end check of the escape hatch.

**Language coverage.** The PR matrix is generated from the diff, so it can include `Go`/`Js`/`Java`/`Yaml` combos the static matrix omits today — a PR touching `aws-go-eks` now gets feedback on it. The nightly matrix is unchanged, so the recurring baseline doesn't suddenly light up red.

## Verification

- `go build` and `go vet -tags all ./...` clean; `shellcheck` clean; both workflows parse and the job graph resolves.
- Detection tested against 13 synthetic diffs in a throwaway worktree, including the third-segment language case (`aws-apigateway-go-routes` → `AwsGo`), `classic-azure-*` → `AzureTs`, deleted examples, top-level-file-only changes, and the shared-path escape hatch.
- Skips confirmed instant (0.00s) with no deploy attempted.
- Table-driven unit tests for the path-matching helper run in the `unit-go` job, so a regression in the scoping mechanism fails fast rather than inside a provider job.

Try it locally with `make changed_examples`.

## Follow-ups

Filed separately: three empty jobs in the nightly matrix, and the fork-PR slash command that has never worked.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -6,24 +6,40 @@ env:
   ESC_ACTION_ENVIRONMENT: imports/github-secrets
   ESC_ACTION_EXPORT_ENVIRONMENT_VARIABLES: false
   PULUMI_ENABLE_JOURNALING: "true"
-
-# The full sweep across every example, so that examples no pull request has touched don't
-# silently rot. Pull requests are handled by pull-request.yml, which runs the same
-# integration tests scoped to just the examples the PR changed.
-#
-# This is also the path for fork pull requests, which can't read ESC/OIDC secrets: a
-# maintainer comments on the PR, command-dispatch.yml turns that into the repository_dispatch
-# below, and the whole suite runs.
-name: Test examples (full sweep)
+name: Pull request
 on:
-  schedule:
-    - cron: 0 9 * * *
-  repository_dispatch:
-    types:
-      - run-tests-command
-  workflow_dispatch: {}
+  pull_request:
+    branches:
+      - master
+
+# Deliberately no `concurrency: cancel-in-progress`. Cancelling a job mid-`pulumi up`
+# orphans real cloud resources with no `pulumi destroy` to clean them up.
 
 jobs:
+  # Works out which examples this PR touched, and which integration test sets need to run as
+  # a result. Everything the PR did not touch is left to the nightly sweep in
+  # test-examples.yml. Needs no secrets, so it finishes in seconds.
+  changes:
+    name: Detect changed examples
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      changed-paths: ${{ steps.changed.outputs.changed-paths }}
+      test-sets: ${{ steps.changed.outputs.test-sets }}
+      run-kubernetes: ${{ steps.changed.outputs.run-kubernetes }}
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          # The script diffs against the PR base, which needs more than a shallow checkout.
+          fetch-depth: 0
+
+      - name: Detect changed examples
+        id: changed
+        run: bash scripts/changed-examples.sh
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+
   lint-ts:
     name: TypeScript lint checks
     runs-on: ubuntu-latest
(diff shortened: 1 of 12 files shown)


jkodroff:
This mirrors what we do in pulumi/docs as far as running a bash script to only run the changed examples.

cnunciato (approved):

Issue #2985, opened 2026-08-21
Upgrade GCP examples to pulumi-gcp v9


pose:
The GCP examples in this repo are a major version behind. `pulumi-gcp` is on 9.x, but almost every GCP example still pins 8.x, and the Java one is older still. These examples are what customers copy, and they are also the source for the registry how-to-guides via `tools/mktutorial`, so the drift is user-visible.

Getting to v9 now is worth doing on its own merits, and it also de-risks the upcoming v10 upgrade: if these examples are still on v8 when v10 lands, every failure has to be triaged as either an 8-to-9 break or a 9-to-10 break before it can be acted on. Related: pulumi/pulumi-gcp#3931.

`gcp-ts-vm-multi-nic` is already on 9.x and needs no work.

### TypeScript

- [x] `gcp-ts-agent-sandbox`
- [x] `gcp-ts-cloudrun`
- [x] `gcp-ts-docker-gcr-cloudrun`
- [x] `gcp-ts-functions`
- [x] `gcp-ts-gke`
- [x] `gcp-ts-gke-hello-world`
- [x] `gcp-ts-gke-serviceaccount`
- [x] `gcp-ts-k8s-ruby-on-rails-postgresql`
- [x] `gcp-ts-oidc-provider-pulumi-cloud`
- [x] `gcp-ts-serverless-raw`
- [x] `gcp-ts-slackbot`
- [x] `kubernetes-ts-multicloud`
- [x] `multicloud-ts-buckets`
- [x] `policy-packs/gcp-ts`

### Python

- [x] `gcp-py-cloudrun-cloudsql`
- [x] `gcp-py-functions`
- [x] `gcp-py-gke`
- [x] `gcp-py-instance-nginx`
- [x] `gcp-py-network-component`
- [x] `gcp-py-serverless-raw`
- [x] `gcp-py-webserver`
- [x] `policy-packs/gcp-python`

### Go

- [x] `gcp-go-functions`
- [x] `gcp-go-functions-raw`
- [x] `gcp-go-gke`
- [x] `gcp-go-instance`
- [x] `gcp-go-webserver`

### .NET

- [x] `gcp-cs-functions`
- [x] `gcp-cs-gke`

### Java

- [x] `gcp-java-gke-hello-world` — its `pom.xml` version range is stale by two majors, not one

### Other

- [x] `secrets-provider/gcloud`

## Acceptance criteria

- [x] Every example above pins pulumi-gcp 9.x, in whichever manifest its language uses.
[...]

PR #2991, opened 2026-09-01
Upgrade the GCP examples to pulumi-gcp v9

pose:
Upgrades the GCP examples to pulumi-gcp v9, so the pending v10 move is a single-major bump rather than a two-major jump.

No source change was required by the 8-to-9 break in any language. The only code edits are the Go import paths (`sdk/v8` -> `sdk/v9`); everything else is a manifest pin, so there is nothing to cross-check against the v9 migration guide.

The Go modules also move `pulumi/sdk/v3` from 3.223.0 to 3.256.0 and the `go` directive from 1.25.0 to 1.25.11. Both are forced rather than chosen: `pulumi-gcp/sdk/v9 v9.35.1` declares `go 1.25.11` and requires `pulumi/sdk/v3 v3.256.0`.

`gcp-hcl-webserver` needs no change: it pins the upstream Terraform provider (`google ~> 7.0`), which already corresponds to pulumi-gcp v9. It will need `~> 8.0` for v10.

### Verification

`pulumi up` needs credentials, so the integration tests run in CI. Locally: all 5 Go examples compile against v9; both .NET examples build with NuGet resolving Pulumi.Gcp 9.35.1; the Python pin resolves to pulumi-gcp 9.35.1 and every `pulumi_gcp` symbol the examples reference still exists in it; and all 16 TypeScript examples were typechecked.

The Java example is not built here: there is no Maven in the toolchain and no integration test covers it. `com.pulumi:gcp` 9.35.1 is on Maven Central and all 9 classes `App.java` imports are present in that jar.

### Pre-existing TypeScript failures, untouched

8 of the 16 TypeScript examples fail to typecheck. Every failure reproduces identically against the v8 pins on `master`, so none is caused by this upgrade, and none of the 8 has integration test coverage, which is how they were able to rot. Left alone here rather than mixed into a version bump.

| Example | Error |
|---|---|
| `gcp-ts-gke-serviceaccount` | `gcp.serviceAccount` should be `gcp.serviceaccount`; `IAMBindingArgs.project` now required |
| `gcp-ts-docker-gcr-cloudrun/cloud-run-deploy` | `Service.status` should be `statuses` |
| `multicloud-ts-buckets` | `{}` not assignable to `BucketArgs` |
| `gcp-ts-gke-hello-world` | `tsconfig.json` lists `cluster.ts` and `config.ts`, which do not exist |
| `gcp-ts-docker-gcr-cloudrun/docker-build-push-gcr` | `Image.digest` gone from `@pulumi/docker` |
| `gcp-ts-k8s-ruby-on-rails-postgresql/infra` | docker `build` typed as `string` |
| `gcp-ts-slackbot` | expression not callable |
| `kubernetes-ts-multicloud` | `deployDashboard` gone from eks `ClusterArgs` |

Part of #2985


jkodroff (approved):
Should be fine as long as CI passes.

PR #3012, opened 2026-09-11
Fail the test when the HTTP retry helper times out

pose:
`misc/test/helpers/http.go`'s retry loop printed a message and returned `false` on timeout without ever touching `*testing.T`. No caller reads that return value, so an example whose endpoint never came up still reported PASS. That is around 46 call sites across the AWS, Google, Kubernetes, DigitalOcean and performance suites.

The timeout path now calls `assert.Failf` and reports what the last attempt actually hit: a transport error, an unexpected status, or a body `ready` had not accepted. Previously `err` was `nil` on any non-200, so the log printed `Http Error: <nil>` and dropped the status code.

**This will turn currently-green tests red**, at least `TestAccGcpHclWebserver` and `TestAccAwsTsApiGateway`. They are already failing and reporting success.

Added `misc/test/helpers/http_test.go`: the three timeout reasons plus the success path, against local `httptest` servers. Runs in 0.047s, no credentials.

### DIFF
--- a/misc/test/helpers/http.go
+++ b/misc/test/helpers/http.go
@@ -32,6 +32,11 @@ func AssertHTTPResultShapeWithRetry(t *testing.T, output interface{}, headers ma
 
 	startTime := time.Now()
 	count, sleep := 0, 0
+	// Why the most recent attempt did not satisfy the check. A transport error, a
+	// non-200 status and a body that `ready` rejected are distinct failure modes,
+	// and the last one seen is the only useful thing to report if we run out of
+	// time, so keep it rather than reporting a bare "no successful GET".
+	lastFailure := "no request was attempted"
 	for {
 		now := time.Now()
 		req, err := http.NewRequest("GET", hostname, nil)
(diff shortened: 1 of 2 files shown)


misc/test/helpers/http.go
```diff
@@ -70,9 +80,14 @@ func AssertHTTPResultShapeWithRetry(t *testing.T, output interface{}, headers ma
 				// Verify it matches expectations
 				return check(bodyText)
 			}
+			lastFailure = "got 200 but the body was not ready for assertion yet"
 		}
 		if now.Sub(startTime) >= maxWait {
-			fmt.Printf("Timeout after %v. Unable to http.get %v successfully.", maxWait, hostname)
+			// No caller reads the returned bool, so the test only fails if we fail
+			// it here. Returning false alone lets a never-successful check pass.
```
iwahbe, line 87: Callers read the returned bool for control flow reasons, just like `assert.Equal`. As a general rule, `Assert*` test helpers should fail the `*testing.T` exactly when they return `false`.

misc/test/helpers/http.go
```diff
@@ -70,9 +80,14 @@ func AssertHTTPResultShapeWithRetry(t *testing.T, output interface{}, headers ma
 				// Verify it matches expectations
 				return check(bodyText)
```
iwahbe, line 81:
```suggestion
				return assert.True(t, check(bodyText), "check failed")
```

misc/test/helpers/http.go
iwahbe, line 25: While we are here...
```suggestion
	if !assert.Truef(t, ok, "expected `%s` output", output) {
```

iwahbe (approved):
If the tests were consistently failing before, you can skip them here. Otherwise we need to wait until they are fixed before merging. 

I don't want to make a red CI run meaningless.

PR #3018, opened 2026-09-11
Migrate docker-build-push-gcr to @pulumi/docker-build

pose:
`gcp-ts-docker-gcr-cloudrun/docker-build-push-gcr` used `docker.Image` from `@pulumi/docker`. It now uses `dockerbuild.Image` from `@pulumi/docker-build`: `imageName` becomes `tags`, `build.context` becomes `context.location`, and `push: true` replaces the implicit push.

The stack output changes shape. `repoDigest` was a fully-qualified `tag@sha256:...`; `digest` is now the bare digest, and `ref` carries the fully-qualified form. Both are exported. No in-repo consumer reads either: the sibling `cloud-run-deploy` project rediscovers the image from the registry by tag rather than through a `StackReference`.

`buildOnPreview` is left unset, matching the other `docker-build` examples in this repo.

Verified with `npm install`, `npm run typecheck` and `make lint_ts`, all clean, plus a `pulumi preview` planning 2 resources. A `typecheck` script was added to `package.json`, which had none.

### DIFF
--- a/gcp-ts-docker-gcr-cloudrun/README.md
+++ b/gcp-ts-docker-gcr-cloudrun/README.md
@@ -55,12 +55,13 @@ This example is split into two Pulumi projects, run in sequence:
     ...
 
     Updating (dev):
-        Type                   Name                 Status
-    +   pulumi:pulumi:Stack    gcr-build-image-dev  created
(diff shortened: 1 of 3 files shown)


guineveresaenger (approved):
nice!

PR #3019, opened 2026-09-14
Run the Python examples CI on Python 3.14

### LINKED ISSUE #3010: Examples CI defaults to Python 3.9
`.github/actions/setup/action.yml` defaults `python-version` to `3.9` and the integration job passes no override, so every Python example is tested on it. Python 3.9 reached end of life in October 2025.og

It now blocks dependency resolution for recent provider SDKs. `pulumi-gcp` 10.x requires `pulumi >= 3.231.0`, and no release in that range publishes a 3.9 wheel:

```
ERROR: Cannot install -r requirements.txt and pulumi<4.0.0 and >=3.5.1 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible
```

Examples pinning a version range avoid this today, because pip backtracks to an older provider release. An exact pin fails immediately.

pulumi/examples#3008 scopes the GCP leg to 3.12 as an unblock. `AwsPy`, `AzurePy` and `DigitalOceanPy` remain on the default. The examples declare no Python version of their own, so the action default is the only constraint.

iwahbe:
The setup action defaulted to Python 3.9, which reached end of life in October 2025. Recent provider SDKs require a pulumi release with no 3.9 wheel, so pip cannot resolve their requirements on the default. Move the default to 3.14 and drop the GcpPy-only 3.12 override, since the default now covers it.

Python 3.14 installs black 26, which applies the 2026 stable style. The reformatted files are the output of `make format` on the new black.

Fixes https://github.com/pulumi/examples/issues/3010


iwahbe:
Generically, examples shouldn't be on our lowest supported version, but our highest. We are _willing_ to support any language version still supported, but we _want_ our users to start on the newest version.

tgummerer (approved):

pose:
Worth getting @julienp thoughts here as well

